PR #9, opened 2025-01-04
chore: add gitattributes so the gh stats are fixed :)

The-personified-devil:
Fixes the github programming language percentage thingy. Bit useless, but why not and I checked that it works on the fork.
Doesn't impact any other git functionality, the `linguist-vendor` attribute is only used for code stats.


Supreeeme:
That's pretty neat :) thanks

PR #16, opened 2025-01-04
fix case when >2 actions use same localized_name

galister:
this fixes vrchat with index controller


galister:
just saw your pr as well so feel free to close this.

Supreeeme:
Can you confirm mine works? I'd prefer to go with my approach but I appreciate the effort :)

PR #29, opened 2025-01-06
input: expose RenderModelName_String property

### LINKED ISSUE #28: Controllers do not work in Resonite
Resonite successfuly boots up into the local home, however controllers don't get recognized and the avatar hands are stuck in the desktop position, none of the bindings work.

Both the resonite and unity log files report various errors, nothing specific in xrizer.txt. 
[xrizer.txt](https://github.com/user-attachments/files/18321603/xrizer.txt)
[Player.log](https://github.com/user-attachments/files/18321604/Player.log)
[ARCHLINUX - 2024.12.26.1378 - 2025-01-06 16_34_14.log](https://github.com/user-attachments/files/18321605/ARCHLINUX.-.2024.12.26.1378.-.2025-01-06.16_34_14.log)

OrionMoonclaw:
fixes #28

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1103,6 +1103,7 @@ impl<C: openxr_data::Compositor> Input<C> {
         struct ProfileData {
             controller_type: &'static CStr,
             model_number: &'static CStr,
+            render_model_name: &'static CStr,
         }
         static PROFILE_MAP: OnceLock<HashMap<xr::Path, ProfileData>> = OnceLock::new();
         let get_profile_data = || {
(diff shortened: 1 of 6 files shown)


Supreeeme:
Does the name itself here actually matter? Could we just make it `xrizer_controller` or something?

OrionMoonclaw:
It does sadly, I tried giving it some random value at first just to see if that would fix it, which caused some bindings not to work, so I'm guessing they're actually using it to detect which controller you're using

Supreeeme:
Do you know if this works correctly for other controllers as well? I wonder if we need to pull the real names, i.e. is resonite just searching for "oculus" in the string and it happens to work.

OrionMoonclaw:
Oh yeah you're right, I got a bit confused by the code in OC I was referencing. Even if all games were just checking substrings it still wouldn't work then since knuckles have names like "valve_controller_knu_1_0_left" for instance.

It's weird that games even depend on this considering there's a dedicated property for the controller type that is even well-documented on the OpenVR wiki, but maybe I shouldn't be surprised :/

PR #31, opened 2025-01-06
input: convert hand tracking data to model space if needed

### LINKED ISSUE #4: Model space skeletal bone data is incorrect
See comment in `input/skeletal.rs`. c331744ee41e741f70ad79bf96224748f4163edb

OrionMoonclaw:
fixes #4

So far I was only able to test this in Resonite and it works there, but Resonite doesn't make use of bone positions (for most of the bones anyway) and aux bones. HL:A crashes for me, so I'm unable to test it, but besides this it will require reference transforms to work correctly anyway.

### DIFF
--- a/src/input/skeletal.rs
+++ b/src/input/skeletal.rs
@@ -24,12 +24,6 @@ impl<C: openxr_data::Compositor> Input<C> {
     ) {
         use HandSkeletonBone::*;
 
-        // TODO: this stuff does not work correctly for model space right now (Alyx)
-        if space == vr::EVRSkeletalTransformSpace::Model {
-            self.get_estimated_bones(session_data, space, hand, transforms);
-            return;
-        }
-
         let legacy = session_data.input_data.legacy_actions.get().unwrap();
         let display_time = self.openxr.display_time.get();
         let Some(raw) = match hand {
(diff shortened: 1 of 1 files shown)


src/input/skeletal.rs
```diff
@@ -145,6 +139,19 @@ impl<C: openxr_data::Compositor> Input<C> {
         {
             xr_joint_to_vr_bone(&joints[joint], &mut transforms[bone as usize])
         }
+
+        // Convert back to model space if needed
+        // it is unnecessary to convert back and forth, but it works and it's easy
+        if space == vr::EVRSkeletalTransformSpace::Model {
+            let bone_data: Vec<(Vec3, Quat)> = parent_to_model_space_bone_data(
+                transforms.iter().map(|t| bone_transform_to_glam(*t))
+            ).collect();
```
Supreeeme, line 148: `collect` should be unnecessary here.

OrionMoonclaw: Borrow checker complains about it because the for loop borrows a mutable reference to transforms, which is still borrowed by the iterator here if the collect is removed :/

Supreeeme:
Will merge after the conflict is fixed.

PR #34, opened 2025-01-11
input: guess skeletal tracking level

OrionMoonclaw:
Makes the skeletal tracking level full if getting bones from hand tracking succeeds, unless the controllers are knuckles.

This allows VRChat to enable hand tracking gestures automatically.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -47,6 +47,7 @@ pub struct Input<C: openxr_data::Compositor> {
     loaded_actions_path: OnceLock<PathBuf>,
     cached_poses: Mutex<CachedSpaces>,
     legacy_packet_num: AtomicU32,
+    skeletal_tracking_level: RwLock<vr::EVRSkeletalTrackingLevel>,
 }
 
 impl<C: openxr_data::Compositor> Input<C> {
@@ -65,6 +66,7 @@ impl<C: openxr_data::Compositor> Input<C> {
             right_hand_key,
             cached_poses: Mutex::default(),
             legacy_packet_num: 0.into(),
+            skeletal_tracking_level: RwLock::new(vr::EVRSkeletalTrackingLevel::Estimated),
         }
     }
 
@@ -372,6 +374,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
         let ActionData::Skeleton { hand, hand_tracker } = action else {
             return vr::EVRInputError::WrongType;
         };
+
         if let Some(hand_tracker) = hand_tracker.as_ref() {
             self.get_bones_from_hand_tracking(
                 &self.openxr,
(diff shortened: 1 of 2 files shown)


src/input.rs
```diff
@@ -380,20 +385,44 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
...
+        let data = self.openxr.session_data.get();
+        let Some(loaded) = data.input_data.get_loaded_actions() else {
+            return vr::EVRInputError::InvalidHandle;
+        };
+
+        let hand = match loaded.try_get_action(action) {
+            Ok(ActionData::Skeleton { hand, .. }) => hand,
+            Ok(_) => return vr::EVRInputError::WrongType,
+            Err(e) => return e,
+        };
```
Supreeeme, line 411:
```suggestion
        get_action_from_handle!(self, action, data, action);
        let ActionData::Skeleton { hand, .. } = action else {
            return vr::EVRInputError::WrongType;
        };
```

src/input.rs
```diff
@@ -380,20 +385,44 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
                 hand_tracker,
                 *hand,
                 transforms,
+                &mut tracking_level,
             )
         } else {
+            *tracking_level = vr::EVRSkeletalTrackingLevel::Estimated;
```
Supreeeme, line 391: I would just move setting this inside of `get_estimated_bones`, instead of having it in 3 different places

src/input/skeletal.rs
```diff
@@ -21,6 +21,7 @@ impl<C: openxr_data::Compositor> Input<C> {
         hand_tracker: &xr::HandTracker,
         hand: Hand,
         transforms: &mut [vr::VRBoneTransform_t],
+        tracking_level: &mut vr::EVRSkeletalTrackingLevel,
```
Supreeeme, line 24: The tracking level is a member of the Input struct, why is it passed as a function parameter?

src/input/skeletal.rs
```diff
@@ -37,11 +38,13 @@ impl<C: openxr_data::Compositor> Input<C> {
...
         .try_get_or_init_raw(xr_data, session_data, &legacy.actions, display_time) else {
+            *tracking_level = vr::EVRSkeletalTrackingLevel::Estimated;
             self.get_estimated_bones(session_data, space, hand, transforms);
             return;
         };
 
         let Some(joints) = raw.locate_hand_joints(hand_tracker, display_time).unwrap() else {
+            *tracking_level = vr::EVRSkeletalTrackingLevel::Estimated;
             self.get_estimated_bones(session_data, space, hand, transforms);
             return;
```
Supreeeme, line 49: If locating joints ever fails, it seems like the tracking level will be permanently stuck as Estimated.

src/input.rs
```diff
@@ -47,6 +47,7 @@ pub struct Input<C: openxr_data::Compositor> {
     loaded_actions_path: OnceLock<PathBuf>,
     cached_poses: Mutex<CachedSpaces>,
     legacy_packet_num: AtomicU32,
+    skeletal_tracking_level: RwLock<vr::EVRSkeletalTrackingLevel>,
```
Supreeeme, line 50: This could probably live in an `AtomicU32` instead, but not a big deal

Supreeeme (commented):
This could use some testing, but I won't block this PR on that.

src/input/skeletal.rs
OrionMoonclaw: rebase fuckup, thanks for catching!

src/input.rs
OrionMoonclaw: I don't really see how that'd fit in with the type system without running additional checks considering this is an enum, so I think RwLock is cleaner

PR #37, opened 2025-01-22
Partially implement stubbed functions for Vivecraft support

cominixo:
This adds basic implementations to application and rendermodel related functions, which are required for Vivecraft to work.
These should probably be properly implemented in the future, but for now this seems fine to get something functional.

### DIFF
--- a/src/applications.rs
+++ b/src/applications.rs
@@ -133,9 +133,11 @@ impl vr::IVRApplications007_Interface for Applications {
         todo!()
     }
     fn GetApplicationCount(&self) -> u32 {
+        crate::warn_unimplemented!("GetApplicationCount");
         0
     }
     fn IsApplicationInstalled(&self, _: *const c_char) -> bool {
+        crate::warn_unimplemented!("IsApplicationInstalled");
         false
     }
     fn RemoveApplicationManifest(&self, _: *const c_char) -> vr::EVRApplicationError {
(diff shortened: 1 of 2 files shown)


OrionMoonclaw:
Would you be down for throwing GetComponentCount in there as well to make VRC work?

Also you can use the crate::warn_unimplemented! macro to print warnings once

cominixo:
~Added GetComponentCount but VRC seems to crash for some other reason to me, throws InvalidParam when creating an action set~

Reverted 50565f4 and VRC seems to be working

src/applications.rs
```diff
@@ -132,15 +133,17 @@ impl vr::IVRApplications007_Interface for Applications {
         todo!()
     }
     fn GetApplicationCount(&self) -> u32 {
-        todo!()
+        0
```
Supreeeme, line 136:
```suggestion
        crate::warn_unimplemented("GetApplicationCount");
        0
```

src/applications.rs
```diff
@@ -132,15 +133,17 @@ impl vr::IVRApplications007_Interface for Applications {
...
     }
     fn IsApplicationInstalled(&self, _: *const c_char) -> bool {
-        todo!()
+        false
```
Supreeeme, line 139:
```suggestion
        crate::warn_unimplemented("IsApplicationInstalled");
        false
```

PR #39, opened 2025-01-23
misc_unknown: fix alyx assert in undoc3

OrionMoonclaw:
Perhaps it's giving us a null pointer?


Supreeeme:
Could you instead check if a and b are null and have them be options? i.e.:
```rust
let a = if a.is_null() { 
  None
} else {
  Some(unsafe { CStr::from_ptr(a) })
};
```

OrionMoonclaw:
That does indeed work

PR #40, opened 2025-01-23
Estimated skeleton rework

OrionMoonclaw:
This fixes the estimated hand skeleton being wonky (including with Quest controllers) and enables full control for the fingers with interpolation at a set speed.

So far has been tested with Quest controllers in VRChat, Resonite and HL:A

### DIFF
--- a/src/input/action_manifest.rs
+++ b/src/input/action_manifest.rs
@@ -78,7 +78,6 @@ impl<C: openxr_data::Compositor> Input<C> {
         let legacy = session_data.input_data.legacy_actions.get_or_init(|| {
             LegacyActionData::new(
                 &self.openxr.instance,
-                &session_data.session,
                 self.openxr.left_hand.subaction_path,
                 self.openxr.right_hand.subaction_path,
             )
--- a/src/input/action_manifest.rs
+++ b/src/input/action_manifest.rs
@@ -6,7 +6,10 @@ use super::{
     profiles::{InteractionProfile, PathTranslation, Profiles},
     BoundPoseType, Input,
 };
-use crate::openxr_data::{self, Hand, SessionData};
+use crate::{
+    input::skeletal::SkeletalInputActionData,
+    openxr_data::{self, Hand, SessionData},
+};
 use log::{debug, error, info, trace, warn};
 use openvr as vr;
 use openxr as xr;
@@ -83,6 +86,17 @@ impl<C: openxr_data::Compositor> Input<C> {
             )
         });
 
+        let skeletal_input = session_data
+            .input_data
+            .estimated_skeleton_actions
+            .get_or_init(|| {
+                SkeletalInputActionData::new(
+                    &self.openxr.instance,
+                    self.openxr.left_hand.subaction_path,
+                    self.openxr.right_hand.subaction_path,
+                )
+            });
+
         // See Input::frame_start_update for the explanation of this.
         let info_set = self
             .openxr
@@ -100,12 +114,12 @@ impl<C: openxr_data::Compositor> Input<C> {
             manifest.default_bindings,
             &legacy.actions,
             &info_action,
+            &skeletal_input,
         );
 
         let xr_sets: Vec<_> = sets
             .values()
-            .chain(std::iter::once(&legacy.set))
-            .chain(std::iter::once(&info_set))
+            .chain([&legacy.set, &info_set, &skeletal_input.set].into_iter())
             .collect();
         session_data.session.attach_action_sets(&xr_sets).unwrap();
 
@@ -753,6 +767,7 @@ impl<C: openxr_data::Compositor> Input<C> {
         bindings: Vec<DefaultBindings>,
         legacy_actions: &LegacyActions,
         info_action: &xr::Action<bool>,
+        skeletal_input: &SkeletalInputActionData,
     ) {
         let mut it: Box<dyn Iterator<Item = DefaultBindings>> = Box::new(bindings.into_iter());
         while let Some(DefaultBindings {
@@ -807,6 +822,7 @@ impl<C: openxr_data::Compositor> Input<C> {
                                 actions,
                                 legacy_actions,
                                 info_action,
+                                skeletal_input,
                                 bindings,
                             );
                         }
@@ -832,6 +848,7 @@ impl<C: openxr_data::Compositor> Input<C> {
         actions: &mut LoadedActionDataMap,
         legacy_actions: &LegacyActions,
         info_action: &xr::Action<bool>,
+        skeletal_input: &SkeletalInputActionData,
         bindings: &HashMap<String, ActionSetBinding>,
     ) {
         info!("loading bindings for {}", profile.profile_path());
@@ -845,6 +862,7 @@ impl<C: openxr_data::Compositor> Input<C> {
         }
         let stp = constrain(|s| self.openxr.instance.string_to_path(s).unwrap());
         let legacy_bindings = profile.legacy_bindings(&stp);
+        let skeletal_bindings = profile.skeletal_input_bindings(&stp);
         let profile_path = stp(profile.profile_path());
         let legal_paths = profile.legal_paths();
         let translate_map = profile.translate_map();
@@ -932,6 +950,7 @@ impl<C: openxr_data::Compositor> Input<C> {
                 info_action,
                 info_action_binding,
             )))
+            .chain(skeletal_bindings.binding_iter(&skeletal_input.actions))
             .collect();
 
         self.openxr
--- a/src/input/legacy.rs
+++ b/src/input/legacy.rs
@@ -1,5 +1,6 @@
 use super::{Input, Profiles};
 use crate::openxr_data::{self, Hand, OpenXrData, SessionData};
+use glam::Quat;
 use log::{debug, trace, warn};
 use openvr as vr;
 use openxr as xr;
@@ -109,12 +110,7 @@ pub(super) struct LegacyActionData {
 }
 
 impl LegacyActionData {
-    pub fn new<'a>(
-        instance: &'a xr::Instance,
-        session: &'a xr::Session<xr::AnyGraphics>,
-        left_hand: xr::Path,
-        right_hand: xr::Path,
-    ) -> Self {
+    pub fn new<'a>(instance: &'a xr::Instance, left_hand: xr::Path, right_hand: xr::Path) -> Self {
         debug!("creating legacy actions");
         let leftright = [left_hand, right_hand];
         let set = instance
@@ -143,12 +139,6 @@ impl LegacyActionData {
             HandSpaces {
                 hand,
                 hand_path,
-                grip: grip_pose
-                    .create_space(session, hand_path, xr::Posef::IDENTITY)
-                    .unwrap(),
-                aim: aim_pose
-                    .create_space(session, hand_path, xr::Posef::IDENTITY)
-                    .unwrap(),
                 raw: OnceLock::new(),
             }
         };
@@ -207,14 +197,8 @@ pub fn setup_legacy_bindings(
 pub(super) struct HandSpaces {
     hand: Hand,
     hand_path: xr::Path,
-    grip: xr::Space,
-    aim: xr::Space,
 
     /// Based on the controller jsons in SteamVR, the "raw" pose
-    /// (which seems to be equivalent to the pose returned by WaitGetPoses)
-    /// is actually the grip pose, but in the same position as the aim pose.
-    /// Using this pose instead of the grip fixes strange controller rotation in
-    /// I Expect You To Die 3.
     /// This is stored as a space so we can locate hand joints relative to it for skeletal data.
     raw: OnceLock<xr::Space>,
 }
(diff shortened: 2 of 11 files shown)


src/input/skeletal.rs
```diff
@@ -395,6 +445,70 @@ static AUX_BONES: &[(HandSkeletonBone, xr::HandJoint)] = &[
...
+            index: 0.0,
+            middle: 0.0,
+            ring: 0.0,
+            pinky: 0.0,
+            thumb: 0.0,
+            time: Instant::now(),
+        }
+    }
+
+    fn lerp(&self, target: &Self, alpha: f32) -> Self {
```
Supreeeme, line 470: "alpha" is a very nondescript name to me. I'd switch it to something like "amount".

Supreeeme (commented):
I haven't given this a full look through yet, but my initial thoughts are I am hesitant to just pull the values from the SteamVR files for the pose offsets. I know for a fact that they are wrong for the Vive wands, and they might be wrong for alternate tracking implementations (i.e. libsurvive). I don't know of a better solution though, worst case scenario I'd just have the SteamVR based transforms as default and allow the user to adjust it themselves at runtime. I will need to think on this.

OrionMoonclaw:
I'll have to look into it again, but I think using the palm_ext pose would be a more portable solution (it has also been promoted to core OpenXR 1.1), we'd have to offset the wrist pose based on that. Having some kind of config would certainly be nice though.

Supreeeme:
If you want to split the commit that adds the SteamVR poses into a different PR I can look at the rest of the commits.

OrionMoonclaw:
Just the one that affects the raw pose, or the skeleton poses too?

Supreeeme:
Either, doesn't matter

src/input/profiles/knuckles.rs
```diff
@@ -96,15 +95,35 @@ impl InteractionProfile for Knuckles {
...
+                    z: 0.005040,
+                    w: 0.990821,
+                },
+                position: xr::Vector3f {
+                    x: 0.004619,
+                    y: -0.020021,
+                    z: -0.129239,
+                },
+            },
+        }
```
Supreeeme, line 139: I'd prefer if these weren't pre-converted, just for the sake of clarity. i.e.:
```rust
let mat = Mat4::from_rotation_translation(
    Quat::from_euler(EulerRot::XYZ, x, y, z),
    Vec3::new(x, y, z)
).inverse();

```

src/input/skeletal_input.rs
```diff
@@ -0,0 +1,72 @@
...
+    thumb_touch: bool,
+    index_touch: bool,
+    index_curl: f32,
+    rest_curl: f32,
+}
+
+pub struct SkeletalInputActionData {
+    pub set: xr::ActionSet,
+    pub actions: SkeletalInputActions,
+}
```
Supreeeme, line 40: Why create a separate set of actions for this?

src/input/skeletal.rs
```diff
@@ -250,6 +231,75 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+        let current_time = std::time::Instant::now();
+
+        let target = FingerState {
+            index,
+            middle: input.rest_curl.max(index / 2.0),
+            ring: input.rest_curl.max(index / 4.0),
+            pinky: input.rest_curl.max(index / 6.0),
+            thumb: if input.thumb_touch { 1.0 } else { 0.0 },
+            time: current_time,
+        };
```
Supreeeme, line 293: What's up with these arbitrary divisions?

Supreeeme (commented):
Since I don't have a better solution at the moment, I'll just accept using the SteamVR transforms, and down the line if we need to we can just allow the user to modify the offsets through some kind of config.

src/input/profiles/vive_controller.rs
```diff
@@ -78,6 +79,35 @@ impl InteractionProfile for ViveWands {
...
+                z: 0.000000,
+                w: 0.999034,
+            },
+            position: xr::Vector3f {
+                x: 0.000000,
+                y: 0.006426,
+                z: -0.097942,
+            },
+        }
+    }
```
Supreeeme, line 110: These offsets seem to make the Wands feel a little further forward than they should. Given that they already feel serviceable enough without any offset, I'd just remove this for now.

src/input/skeletal.rs
OrionMoonclaw: It makes the other fingers curl a bit together with the index to follow human anatomy (and SteamVR), the values are just whatever felt right.

src/input/skeletal_input.rs
Supreeeme: Fair enough. I think these can go inside of `skeletal.rs` though, no need for a separate file.

src/input/skeletal.rs
Supreeeme: Can that be noted with a comment somewhere?

OrionMoonclaw:
I ran into some kind of math error for now, won't have time to fix today so tomorrow it is, sorry for the delay

OrionMoonclaw:
Radians, of course...

Should be good now

src/input/legacy.rs
```diff
@@ -207,12 +197,8 @@ pub fn setup_legacy_bindings(
...
     hand: Hand,
     hand_path: xr::Path,
-    grip: xr::Space,
-    aim: xr::Space,
 
     /// Based on the controller jsons in SteamVR, the "raw" pose
-    /// (which seems to be equivalent to the pose returned by WaitGetPoses)
-    /// is actually the grip pose, but in the same position as the aim pose.
     /// Using this pose instead of the grip fixes strange controller rotation in
     /// I Expect You To Die 3.
```
Supreeeme, line 203: This comment can go too

src/input/profiles/vive_controller.rs
```diff
@@ -76,6 +81,23 @@ impl InteractionProfile for ViveWands {
             squeeze: stp.leftright("input/squeeze/click"),
         }
     }
+
+    fn skeletal_input_bindings(&self, stp: &dyn StringToPath) -> SkeletalInputBindings {
+        SkeletalInputBindings {
+            thumb_touch: stp
+                .leftright("input/trigger/click")
```
Supreeeme, line 88: Trigger click for thumb touch? Maybe meant to be `input/trackpad/click`?

src/input/skeletal.rs
```diff
@@ -247,11 +228,133 @@ impl<C: openxr_data::Compositor> Input<C> {
...
         }
 
-        // TODO: This is an arbitrary transform, and only appears to work as expected in parent space.
-        transforms[Root as usize].position = Vec3::new(0.0, 0.0, -0.15).into();
-
         *self.skeletal_tracking_level.write().unwrap() = vr::EVRSkeletalTrackingLevel::Estimated;
     }
+
+    fn get_finger_state(&self, session_data: &SessionData, hand: Hand) -> FingerState {
+        let data = self.openxr.session_data.get();
```
Supreeeme, line 235: Should use the passed in SessionData here

src/input/skeletal.rs
```diff
@@ -247,11 +228,133 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+            .input_data
+            .estimated_skeleton_actions
+            .get()
+            .unwrap()
+            .actions;
+        let subaction = match hand {
+            Hand::Left => self.openxr.left_hand.subaction_path,
+            Hand::Right => self.openxr.right_hand.subaction_path,
+        };
+        let input = SkeletalInputActionStates {
```
Supreeeme, line 246: What is the purpose of this struct? It's only ever used here.

src/input/skeletal.rs
```diff
@@ -247,11 +228,133 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+            .max(if input.index_touch || input.index_curl > 0.0 {
+                0.3
+            } else {
+                0.0
+            });
+
+        let mut state = self.estimated_finger_state[match hand {
+            Hand::Left => 0,
+            Hand::Right => 1,
+        }]
```
Supreeeme, line 280:
```suggestion
        let mut state = self.estimated_finger_state[hand as usize - 1]
```

src/input/skeletal.rs
```diff
@@ -247,11 +228,133 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+            // Make other fingers curl with the index slightly to mimic how real human hands work
+            middle: input.rest_curl.max(index / 2.0),
+            ring: input.rest_curl.max(index / 4.0),
+            pinky: input.rest_curl.max(index / 6.0),
+            thumb: if input.thumb_touch { 1.0 } else { 0.0 },
+            time: current_time,
+        };
+
+        let elapsed_time = current_time.duration_since(state.time).as_secs_f32();
+        let alpha = (elapsed_time * 24.0).min(1.0);
```
Supreeeme, line 297: Need comment, or preferably a `const` explaing why `* 24.0`

Supreeeme, line 297:
```suggestion
        let t = (elapsed_time * 24.0).min(1.0);
```
`alpha` elicits color terminology to me.

src/input/skeletal.rs
```diff
@@ -247,11 +228,133 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+                    orientation: rot.into(),
+                };
+            }
+        }
+
+        match full_it {
+            TransformedIt::Parent(it) => convert(it, transforms),
+            TransformedIt::Model(it) => convert(it, transforms),
+        }
+    }
```
Supreeeme, line 357: This is copy pasted from get_estimated_bones and should be turned into a function

src/input/action_manifest.rs
```diff
@@ -101,12 +114,14 @@ impl<C: openxr_data::Compositor> Input<C> {
...
             &legacy.actions,
             &info_action,
+            &skeletal_input,
         );
 
         let xr_sets: Vec<_> = sets
             .values()
             .chain(std::iter::once(&legacy.set))
             .chain(std::iter::once(&info_set))
+            .chain(std::iter::once(&skeletal_input.set))
```
Supreeeme, line 124:
```suggestion
            .values()
            .chain([&legacy.set, &info_set, &skeletal_input.set].into_iter())
```

src/input/skeletal.rs
OrionMoonclaw: Not entirely certain how to do it with all the type weirdness going on, will need to think about it a bit longer

Supreeeme: How about:
```rust
fn finalize_transforms(bone_iterator: impl PoseIterator, space: vr::EVRSkeletalTransformSpace, trasnforms: &mut [vr::VRBoneTransform_t]) {
// impl
}
```

src/input/skeletal.rs
```diff
@@ -247,11 +228,128 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+                openvr::EVRSkeletalReferencePose::BindPose => &gen::left_hand::BINDPOSE,
+                openvr::EVRSkeletalReferencePose::OpenHand => &gen::left_hand::OPENHAND,
+                openvr::EVRSkeletalReferencePose::Fist => &gen::left_hand::FIST,
+                openvr::EVRSkeletalReferencePose::GripLimit => &gen::left_hand::GRIPLIMIT,
+            },
+            Hand::Right => match pose {
+                openvr::EVRSkeletalReferencePose::BindPose => &gen::right_hand::BINDPOSE,
+                openvr::EVRSkeletalReferencePose::OpenHand => &gen::right_hand::OPENHAND,
+                openvr::EVRSkeletalReferencePose::Fist => &gen::right_hand::FIST,
+                openvr::EVRSkeletalReferencePose::GripLimit => &gen::right_hand::GRIPLIMIT,
```
Supreeeme, line 317:
```suggestion
                vr::EVRSkeletalReferencePose::BindPose => &gen::left_hand::BINDPOSE,
                vr::EVRSkeletalReferencePose::OpenHand => &gen::left_hand::OPENHAND,
                vr::EVRSkeletalReferencePose::Fist => &gen::left_hand::FIST,
                vr::EVRSkeletalReferencePose::GripLimit => &gen::left_hand::GRIPLIMIT,
            },
            Hand::Right => match pose {
                vr::EVRSkeletalReferencePose::BindPose => &gen::right_hand::BINDPOSE,
                vr::EVRSkeletalReferencePose::OpenHand => &gen::right_hand::OPENHAND,
                vr::EVRSkeletalReferencePose::Fist => &gen::right_hand::FIST,
                vr::EVRSkeletalReferencePose::GripLimit => &gen::right_hand::GRIPLIMIT,
```

src/input/skeletal.rs
```diff
@@ -247,11 +228,128 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+            .rest_curl
+            .state(&session_data.session, subaction)
+            .unwrap()
+            .current_state;
+
+        let index = index_curl.max(if index_touch || index_curl > 0.0 {
+            0.3
+        } else {
+            0.0
+        });
```
Supreeeme, line 273: Why 0.3?

src/input/skeletal.rs
```diff
@@ -247,11 +228,128 @@ impl<C: openxr_data::Compositor> Input<C> {
...
         }
 
-        // TODO: This is an arbitrary transform, and only appears to work as expected in parent space.
-        transforms[Root as usize].position = Vec3::new(0.0, 0.0, -0.15).into();
-
         *self.skeletal_tracking_level.write().unwrap() = vr::EVRSkeletalTrackingLevel::Estimated;
     }
+
+    fn get_finger_state(&self, session_data: &SessionData, hand: Hand) -> FingerState {
+        const FINGER_SMOOTHING_SPEED: f32 = 24.0;
```
Supreeeme, line 235: It's still not clear to me what this is doing :sweat_smile: Why are you multiplying it?

Supreeeme (commented):
Some final nitpicks, otherwise looks good

src/input/skeletal.rs
OrionMoonclaw: That works, I'm actually not entirely sure what I screwed up when trying to do it heh

Supreeeme:
Thanks!

PR #41, opened 2025-01-26
Implement action set priorities and restrict to device

OrionMoonclaw:
This splits out skeleton inputs into another process (thanks openxr loader...) with a headless session and implements action set priorities!

Mostly tested for regressions with VRChat (needed device restriction and secondary action sets to work properly), Resonite (doesn't need priorities anyway) and Alyx.

Would appreciate testing with other games and controllers, especially Valheim VR, assuming that works with xrizer at all.

Depends on #40

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -32,6 +32,7 @@ use std::ffi::{CStr, CString, c_char, c_void};
 use std::mem::ManuallyDrop;
 use std::ops::Deref;
 use std::path::PathBuf;
+use std::ptr;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex, OnceLock, RwLock, RwLockReadGuard};
@@ -1012,13 +1013,6 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
         let active_sets =
             unsafe { std::slice::from_raw_parts(active_sets, active_set_count as usize) };
 
-        if active_sets
-            .iter()
-            .any(|set| set.ulRestrictedToDevice != vr::k_ulInvalidInputValueHandle)
-        {
-            crate::warn_once!("Per device action set restriction is not implemented yet.");
-        }
-
         let data = self.openxr.session_data.get();
         let Some(actions) = data.input_data.get_loaded_actions() else {
             return vr::EVRInputError::InvalidParam;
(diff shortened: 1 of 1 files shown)


Supreeeme:
I'd really prefer not to use multiple instances and IPC to accomplish this - that seems error prone, hard to test, and not portable across runtimes. I have some thoughts on how to accomplish this that I'd like to explore first.

OrionMoonclaw:
I think as far as testing it shouldn't be too bad, I just need to make some changes to fakexr to support headless sessions. The IPC interface is very minimal and uses ipc-channel, so not hard to manually verify and at runtime it will just return an error if anything goes wrong. I still need to make sure it doesn't panic when that happens since it's better to not have working fingers than to just outright crash.

I don't think it's too big of an ask for PC XR runtimes to support headless sessions if they want better compatibility with OpenVR games (some of them do already, including Monado and VDXR), XR_EXT_active_action_set_priority would be the bigger issue here with afaik pretty much only Monado and Meta supporting it (VDXR is open source, so support could be added there).

I think what I'm actually least happy about currently is the way the binary is found, I don't think that's portable and it relies on where vrclient.so is located. I tried embedding the binary and using memfd_create, but that's even more Linux-specific and would complicate the build process, perhaps writing it to a temp file instead would work better.

OrionMoonclaw:
Update: my current plan with this is to only enable priorities if there's hand tracking data available from the runtime, [as there seems to be interest on the Monado side for adding skeletal estimation there](https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2637#note_3202222), so it will be a matter of creating a generic API for it and gradually adding support for different devices

I'm still waiting on some of @ImSapphire's PRs that touch hand tracking code to avoid having to rewrite code due to merge conflicts

galister:
@OrionMoonclaw hi, would you mind rebasing the fork on upstream main and maybe activating CI on the fork so that people who need priorities support can easily obtain it? there's no better option right now, as far as I'm aware.

galister:
@Supreeeme would you care to make ghcr.io/supreeeme/xrizer-ubuntu publicly accessible so that we can run CI on the forks?

OrionMoonclaw:
> @OrionMoonclaw hi, would you mind rebasing the fork on upstream main and maybe activating CI on the fork so that people who need priorities support can easily obtain it? there's no better option right now, as far as I'm aware.

I've got some time now so I can definitely rebase it at least, though I've never touched github's CI so we'll see how that goes

PR #49, opened 2025-02-13
Refactor Devices & Implement XR_MNDX_xdev_space

RinLovesYou:
A big one, If you'd like me to split this up into smaller PR's I'm happy to do so

This separates device logic into a trait for different types to implement, similar to how OpenComposite was structured.
This allows for cleaner separation of device specific logic, i.e. getting HMD poses, Controller poses, Tracker poses, and anything else we may add in the future.

the XR_MNDX_xdev_space implementation required me to write our own bindings as it is not part of the OpenXR spec yet(?) and doesn't get generated by the openxrs crate.

Marked as draft while i'm still cleaning up the initial implementation


RinLovesYou:
So many changes happened that i will redo a split of this into two PRs

PR #50, opened 2025-02-15
temporarily load custom bindings like opencomposite

RinLovesYou:
until a better solution is done

### DIFF
--- a/src/input/action_manifest.rs
+++ b/src/input/action_manifest.rs
@@ -15,11 +15,14 @@ use serde::{
     Deserialize,
 };
 use slotmap::SecondaryMap;
-use std::{cell::{LazyCell, RefCell}, env::current_dir};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::RwLock;
+use std::{
+    cell::{LazyCell, RefCell},
+    env::current_dir,
+};
 
 impl<C: openxr_data::Compositor> Input<C> {
     pub(super) fn load_action_manifest(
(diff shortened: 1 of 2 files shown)


Supreeeme:
What's the benefit of this over just editing the bindings that the game comes with?

RinLovesYou:
> What's the benefit of this over just editing the bindings that the game comes with?

EAC games don't like you messing with existing game files, even on proton

openglfreak:
Could the directory or file to load be taken from an environment variable instead of being "hard-coded"?

openglfreak:
It's not the best idea to add files to an AC-protected game either; While currently it seems that EAC doesn't care all that much, this could conceivably change, or maybe another AC takes more issue with it than EAC.

RinLovesYou:
> Could the directory or file to load be taken from an environment variable instead of being "hard-coded"?

Can do

> It's not the best idea to add files to an AC-protected game either; While currently it seems that EAC doesn't care all that much, this could conceivably change, or maybe another AC takes more issue with it than EAC.

I don't think it's a huge concern, anticheat only really cares about existing game files, so we can keep it as a default path that can be overridden?

openglfreak:
Sure!

src/input/action_manifest.rs
```diff
@@ -759,7 +759,18 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+                let custom_path = if let Ok(custom_dir) = std::env::var("XRIZER_CUSTOM_BINDINGS_DIR") {
+                    PathBuf::from(custom_dir)
+                } else {
+                    current_dir().unwrap().join("xrizer")
+                }.join(format!("{controller_type:?}.json").to_lowercase());
+                let bindings_path = if custom_path.exists() {
+                    info!("Using custom bindings for {controller_type:?} (at {})", custom_path.display());
+                    custom_path
+                } else {
+                    info!("Using default bindings for {controller_type:?} (at {})", parent_path.join(&binding_url).display());
```
Supreeeme, line 771: I think this log is unnecessary, it's done right below

PR #56, opened 2025-02-25
fix index grip/click never triggering

galister:
not sure why it works with value, but not with force.

OC also uses value for click.


Supreeeme:
Really, it nevers triggers? Not even if you squeeze as hard as possible?

galister:
In VRC the in-world grab never triggers.

In No Man's Sky, I tested with the cockpit stick and throttle, had to constantly death grip them to work.

galister:
closing as superseded by https://github.com/Supreeeme/xrizer/pull/59

PR #59, opened 2025-02-28
Bindings rework

knah:
This PR includes the following changes to the input system:
* Store custom binding information per-profile and per-hand explicitly
  * Most (all?) bindings in SteamVR are specifying only one hand as input source anyway
  * This incidentally allows detecting action origin for custom bindings, which is used when possible
  * This also prevents cross-contamination of custom bindings between different profiles/controller types.
* Custom binding information and state is stored separately from actions
* Adjusted index controller grip thresholds to be in-line with SteamVR
  * This changes how the "grab" action is read as a `f32` - it's now binary 0/1 and never returns any inbetween values. A common `f32` binding type for grip seems to be "trigger" or "force_sensor", but if you have examples of games relying on "grab" binding type having a `f32` value - please let me know!
* Implemented custom thresholds for "button" bindings
  * This includes support for alternate click source, such as force; default is "value" or "click", in that order
  * This seems to be consistent with binding to clicky triggers - a "button" binding doesn't actually require a click, instead a "trigger" binding type should be used if a click is desired
  * Also implemented "touch" input for "button" bindings
* Implemented force threshold and haptic feedback for "dpad" bindings on index controller touchpad
* Simple controller profile no longer pretends to be a vive wand. I don't expect any game to actually ship with bindings for this profile, although SteamVR binding UI has it in its long controller list.
* Minor fixes for various path translation and binding parsing issues

Games tested by me so far with index controllers:
* VRC - most actions work fine, but gestures fail due to missing skeletal summary data
* H3VR - played without issues, seems to be consistent with OC/SteamVR

More testing with a more varied set of games would be welcome

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -16,7 +16,7 @@ use crate::{
     openxr_data::{self, Hand, OpenXrData, SessionData},
     tracy_span, AtomicF32,
 };
-use custom_bindings::{BoolActionData, FloatActionData};
+use custom_bindings::{BindingData, DpadActions, GrabActions};
 use legacy::{setup_legacy_bindings, LegacyActionData};
 use log::{debug, info, trace, warn};
 use openvr::{self as vr, space_relation_to_openvr_pose};
@@ -139,6 +139,67 @@ impl<C: openxr_data::Compositor> Input<C> {
             }
         }
     }
+
+    fn state_from_bindings_left_right(&self, action: vr::VRActionHandle_t) -> Option<(xr::ActionState<bool>, vr::VRInputValueHandle_t)>{
+        debug_assert!(self.left_hand_key.0.as_ffi() != 0);
+        debug_assert!(self.right_hand_key.0.as_ffi() != 0);
+        let left_state = self.state_from_bindings(action, self.left_hand_key.0.as_ffi());
+
+        match left_state {
+            None => self.state_from_bindings(action, self.right_hand_key.0.as_ffi()),
+            Some((left, _)) => {
+                if left.is_active && left.current_state {
+                    return left_state
+                }
+                let right_state = self.state_from_bindings(action, self.right_hand_key.0.as_ffi());
+                match right_state {
+                    None => left_state,
+                    Some((right, _)) => {
+                        if right.is_active && right.current_state {
+                            return right_state
+                        }
+                        if left.is_active {
+                            return left_state
+                        }
+                        right_state
+                    }
+                }
+            }
+        }
+    }
+
+    fn state_from_bindings(&self, action: vr::VRActionHandle_t, restrict_to_device: vr::VRInputValueHandle_t) -> Option<(xr::ActionState<bool>, vr::VRInputValueHandle_t)> {
+        let subaction = self.subaction_path_from_handle(restrict_to_device)?;
+        if subaction == xr::Path::NULL {
+            return self.state_from_bindings_left_right(action);
+        }
+
+        let session = self.openxr.session_data.get();
+        let Ok(loaded_actions) = session.input_data.loaded_actions.get()?.read() else {
+            return None;
+        };
+
+        let interaction_profile = session.session.current_interaction_profile(subaction).ok()?;
+        let bindings = loaded_actions.try_get_bindings(action, interaction_profile).ok()?;
+        let extra_data = loaded_actions.try_get_extra(action).ok()?;
+
+        let mut best_state: Option<xr::ActionState<bool>> = None;
+
+        for x in bindings.iter() {
+            let Ok(Some(state)) = x.state(&session, extra_data, subaction) else {
+                continue;
+            };
+
+            if state.is_active && (!best_state.is_some_and(|x| x.is_active) || state.current_state && !best_state.is_some_and(|x| x.current_state)) {
+                best_state = Some(state);
+                if state.current_state {
+                    break;
+                }
+            }
+        }
+
+        best_state.map(|x| (x, restrict_to_device))
+    }
 }
 
 #[derive(Default)]
@@ -155,17 +216,16 @@ impl InputSessionData {
     }
 }
 enum ActionData {
-    Bool(BoolActionData),
-    Vector1(FloatActionData),
+    Bool(xr::Action<bool>),
+    Vector1 {
+        action: xr::Action<f32>,
+        last_value: AtomicF32,
+    },
     Vector2 {
         action: xr::Action<xr::Vector2f>,
         last_value: (AtomicF32, AtomicF32),
     },
-    Pose {
-        /// Maps an interaction profile path to whatever kind of pose was bound for this action for
-        /// that profile.
-        bindings: HashMap<xr::Path, BoundPose>,
-    },
+    Pose,
     Skeleton {
         hand: Hand,
         hand_tracker: Option<xr::HandTracker>,
@@ -173,7 +233,14 @@ enum ActionData {
     Haptic(xr::Action<xr::Haptic>),
 }
 
-#[derive(Debug)]
+#[derive(Default)]
+struct ExtraActionData {
+    pub dpad_actions: Option<DpadActions>,
+    pub toggle_action: Option<xr::Action<bool>>,
+    pub grab_action: Option<GrabActions>
+}
+
+#[derive(Debug, Default)]
 struct BoundPose {
     left: Option<BoundPoseType>,
     right: Option<BoundPoseType>,
@@ -190,12 +257,16 @@ enum BoundPoseType {
 
 macro_rules! get_action_from_handle {
     ($self:expr, $handle:expr, $session_data:ident, $action:ident) => {
+        get_action_from_handle!($self, $handle, $session_data, $action, loaded)
+    };
+
+    ($self:expr, $handle:expr, $session_data:ident, $action:ident, $loaded:ident) => {
         let $session_data = $self.openxr.session_data.get();
-        let Some(loaded) = $session_data.input_data.get_loaded_actions() else {
+        let Some($loaded) = $session_data.input_data.get_loaded_actions() else {
             return vr::EVRInputError::InvalidHandle;
         };
 
-        let $action = match loaded.try_get_action($handle) {
+        let $action = match $loaded.try_get_action($handle) {
             Ok(action) => action,
             Err(e) => return e,
         };
@@ -609,7 +680,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
         }
         let subaction_path = get_subaction_path!(self, restrict_to_device, action_data);
         let (active_origin, hand) = match loaded.try_get_action(action) {
-            Ok(ActionData::Pose { bindings }) => {
+            Ok(ActionData::Pose) => {
                 let (mut hand, interaction_profile) = match subaction_path {
                     x if x == self.openxr.left_hand.subaction_path => (
                         Some(Hand::Left),
@@ -624,13 +695,14 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
                 };
 
                 let get_first_bound_hand_profile = || {
-                    bindings
-                        .get(&self.openxr.left_hand.profile_path.load())
-                        .or_else(|| bindings.get(&self.openxr.right_hand.profile_path.load()))
+                    loaded
+                        .try_get_pose(action, self.openxr.left_hand.profile_path.load())
+                        .or_else(|_| loaded.try_get_pose(action, self.openxr.right_hand.profile_path.load()))
+                        .ok()
                 };
 
                 let Some(bound) = interaction_profile
-                    .and_then(|p| bindings.get(&p))
+                    .and_then(|p| loaded.try_get_pose(action, p).ok())
                     .or_else(get_first_bound_hand_profile)
                 else {
                     match hand {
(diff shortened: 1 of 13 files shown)


Supreeeme:
This is a lot of changes, some of which are unrelated, and it would be easier for me to review if it was split into a few separate PRs.

src/input/action_manifest.rs
```diff
@@ -1015,29 +1083,43 @@ fn handle_dpad_binding(
...
+        let mut data = extra_actions.remove(action_name).unwrap_or_default();
+
+        if data.dpad_actions.is_none() {
+            // FIXME: this steals a random action from app's action set for haptic feedback; there might be none
+            let haptic = if matches!(controller_type, ControllerType::Knuckles) && parent_path.ends_with("trackpad") {
+                let actions = actions.borrow();
+                actions.iter().filter_map(|(_, a)| if let ActionData::Haptic(h) = a { Some(h) } else { None }).find(|_| true).map(|x| x.clone())
+            } else {
+                None
+            };
```
Supreeeme, line 1100: I don't like this - just create a haptic action for it.

src/input/action_manifest.rs
```diff
@@ -1015,29 +1083,43 @@ fn handle_dpad_binding(
...
 
-        if data.dpad_data.is_none() {
             let (parent_action, click_or_touch) = LazyCell::force(&created_actions);
-            data.dpad_data = Some(DpadData {
-                parent: parent_action.clone(),
+            data.dpad_actions = Some(DpadActions {
+                direction: parent_action.clone(),
```
Supreeeme, line 1104: Why is this named `direction`?

src/input/action_manifest.rs
```diff
@@ -1148,16 +1238,32 @@ fn translate_warn(action: &str) -> impl FnOnce(&InvalidActionPath) + '_ {
...
+    let hand_prefix = if path.starts_with("/user/hand/left") {
+        "/user/hand/left"
+    } else if path.starts_with("/user/hand/right") {
+        "/user/hand/right"
+    } else {
+        return None;
+    };
+
+    let path = instance.string_to_path(hand_prefix).ok();
+    path.map(|x| if x == xr::Path::NULL { None } else { Some(x) }).flatten()
```
Supreeeme, line 1251:
```suggestion
    path.and_then(|x| if x == xr::Path::NULL { None } else { Some(x) })
```

src/input/action_manifest.rs
```diff
@@ -1204,54 +1310,131 @@ fn handle_sources(
...
+                    let translated = if let Ok(translated) = path_translator(&format!("{path}/{target}"))
+                        .inspect_err(|e| info!("Falling back to click for {output} ({})", e.0)) {
+                        translated
+                    } else {
+                        let Ok(translated) = path_translator(&format!("{path}/click"))
+                            .inspect_err(translate_warn(output)) else {
+                            continue;
+                        };
+                        translated
+                    };
```
Supreeeme, line 1398:
```suggestion
                    } else if let Ok(translated) = path_translator(&format!("{path}/click"))
                            .inspect_err(translate_warn(output))  {
                            translated
                        };
                    } else {
                        continue;
                    };
```

src/input/custom_bindings.rs
```diff
@@ -558,64 +472,54 @@ mod tests {
...
+        value_state_check(0.0, grab + 0.1, true, true, line!());
+        value_state_check(0.1, 0.0, true, false, line!());
         value_state_check(0.0, 1.0, true, false, line!());
         value_state_check(0.0, release, false, true, line!());
-        value_state_check(grab - 0.1, 1.0, false, false, line!());
+        value_state_check(0.0, grab - 0.1, false, false, line!());
     }
 
     #[test]
-    fn grab_per_hand() {
```
Supreeeme, line 571: It seems this test has been completely removed?

src/input/profiles.rs
```diff
@@ -23,6 +23,7 @@ pub trait InteractionProfile: Sync + Send {
     fn offset_grip_pose(&self, pose: xr::Posef) -> xr::Posef {
         pose
     }
+    fn controller_type(&self) -> ControllerType;
```
Supreeeme, line 26: Why add this? This association is already handled by `Profiles::get`.

Supreeeme (changes requested):

src/input.rs
```diff
@@ -185,6 +245,18 @@ macro_rules! get_action_from_handle {
...
+        let $session_data = $self.openxr.session_data.get();
+        let Some($loaded) = $session_data.input_data.get_loaded_actions() else {
+            return vr::EVRInputError::InvalidHandle;
+        };
+
+        let $action = match $loaded.try_get_action($handle) {
+            Ok(action) => action,
+            Err(e) => return e,
+        };
+    };
```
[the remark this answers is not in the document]
knah: The reason for separate branch is exposing `loaded` to the macro user. I replaced the original branch with a use of this one.

src/input.rs
```diff
@@ -125,6 +125,60 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+
+        let mut best_state: Option<xr::ActionState<bool>> = None;
+
+        for x in bindings.iter() {
+            let Ok(Some(state)) = x.state(&session, extra_data, subaction) else {
+                continue;
+            };
+
+            if state.is_active && (!best_state.map(|x| x.is_active).unwrap_or(false) || state.current_state && !best_state.map(|x| x.current_state).unwrap_or(false)) {
+                best_state = Some(state);
```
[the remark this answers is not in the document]
knah: This could have been a leftover from my initial draft where custom bindings were `f32`. Added a `break`.

src/input/action_manifest.rs
```diff
@@ -25,6 +25,21 @@ use std::{
...
+            .map(|(name, action)| {
+                let key = act_guard
+                    .iter()
+                    .find_map(|(key, super::Action { path })| (*path == name).then_some(key))
+                    .unwrap_or_else(|| act_guard.insert(super::Action { path: name }));
+
+                (key, action)
+            })
+            .collect()
+    }
```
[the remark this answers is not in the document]
knah: I'm not quite sure which one - `action_map_to_secondary` uses the lock-protected map from `self`, and the lambda in `map` uses captured variables too.

src/input/action_manifest.rs
knah: My concern here is action set priorities. The spec seemingly says nothing about how action set priorities interact with output actions, but it would be a shame to break game's haptic output in case a dpad haptic from a higher-priority action set does in fact override it.

src/input/action_manifest.rs
knah: A "dpad" binding has two input actions - the direction and the (possibly optional) click. This naming seems to be cleaner, especially in the context of `DpadActions`

src/input/custom_bindings.rs
knah: Oops, it got eaten by the other test, I put it back in.

src/input/profiles.rs
knah: I reverted the commit that includes this and source the `ControllerType` from caller instead.

knah:
> This is a lot of changes, some of which are unrelated, and it would be easier for me to review if it was split into a few separate PRs.

I hope that the separation of changes into individual well-defined commits does make reviewing easier. GH PR UI allows viewing single-commit diffs via a selector in the top-left.


I fixed up a few more occurrences of `map().flatten()` in addition to your suggestion.

Let me know if you want me to `rebase --autosquash` to clean up commit history before merge

src/input/action_manifest.rs
Supreeeme: A direction to me implies an actual direction (north, south, east, west). A "direction" action is unclear, especially considering this is a Vector2f action which doesn't return a specific direction but rather 2 floats.

knah: I'm up for naming it something more descriptive than either "direction" or "parent_action", perhaps something straightforward like "xy"?

Supreeeme: I can live with that.

Supreeeme (approved):
Looks ok so far, just minor things remaining. Tested it in a few of my games and they seemed fine. Hopefully the test coverage is enough. Be sure to run `cargo fmt` and `cargo clippy`.

src/input/action_manifest.rs
```diff
@@ -190,12 +222,14 @@ struct DefaultBindings {
...
 
-#[derive(Deserialize, Debug, PartialEq, Eq)]
+#[derive(Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
 #[serde(rename_all = "snake_case")]
 pub(super) enum ControllerType {
     ViveController,
     Knuckles,
     OculusTouch,
+    // AFAIK there's no bindings for it, but also there's no reason to suggest bindings for it
+    SimpleController,
```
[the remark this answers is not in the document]
knah: I see what happened there - the original approach had maps keyed with `ControllerType` instead of `xr::Path`, so I wanted to have profiles fully separate. I restored the existing logic - although it does spam the log quite a bit with mismatched bindings.

knah:
I squashed the fixups and did both `fmt` and `clippy`. There's leftover warnings about functions having too many arguments that I don't quite believe to be useful - shoving that same data into a `struct` won't help readability I'd say.

There was also a lost dpad haptic due to missing subaction paths on dpad bindings - I fixed that one too.

Supreeeme:
> There's leftover warnings about functions having too many arguments that I don't quite believe to be useful

Add `#[allow(clippy::too_many_arguments)]` to these.
If you don't mind fixing the merge conflict I will merge this - think it's just a test that got removed.

knah:
I did end up splitting some state/code into separate file to hopefully make it easier to read. I also rebased it on latest main.

knah:
Hopefully last changes I'll shove in here, these fix issues seen in Alyx, bringing it to playable on index and touch.

src/input/action_manifest_helpers.rs
```diff
@@ -0,0 +1,604 @@
+use crate::input::action_manifest::{
```
Supreeeme, line 1: I'd prefer if this file were a module under `action_manifest`, structured like this:
```
action_manifest/helpers.rs
action_manifest.rs
```

src/input/action_manifest_helpers.rs
```diff
@@ -0,0 +1,604 @@
...
+            );
+
+            extra_data.vector2_action = Some(float_action);
+        }
+        self.extra_actions.insert(output.to_lowercase(), extra_data);
+
+        float_name_with_as
+    }
+
+    pub fn get_or_create_grip_action_pair(
```
Supreeeme, line 422:
```suggestion
    pub fn get_or_create_grab_action_pair(
```
The type is "grab" in the binding files.

src/input/action_manifest.rs
```diff
@@ -671,11 +694,11 @@ struct DpadInput {
...
-struct DpadParameters {
-    sub_mode: DpadSubMode,
-    deadzone_pct: FromString<u8>,
-    overlap_pct: FromString<u8>,
-    sticky: FromString<bool>,
+pub struct DpadParameters {
+    pub sub_mode: DpadSubMode,
+    pub deadzone_pct: FromString<u8>,
+    pub overlap_pct: FromString<u8>,
+    pub sticky: FromString<bool>,
```
Supreeeme, line 701: If `action_manifest_helpers` is made a submodule then this stuff doesn't have to be made pub.

src/input/action_manifest_helpers.rs
```diff
@@ -0,0 +1,604 @@
...
+                };
+                DpadHapticData {
+                    action: action.clone(),
+                    key: haptic_key,
+                    binding: hand_path,
+                }
+            })
+        } else {
+            None
+        };
```
Supreeeme, line 592:
```rust
use_force.then(|| {
// ...
});
```

src/input/profiles.rs
```diff
@@ -98,6 +98,7 @@ impl Profiles {
                 (ControllerType::Knuckles, &Knuckles),
                 (ControllerType::OculusTouch, &Touch),
                 (ControllerType::ViveController, &SimpleController),
```
Supreeeme, line 100: Did you mean to delete this line?

src/input/action_manifest.rs
```diff
@@ -4,7 +4,7 @@ use super::{
     profiles::{PathTranslation, Profiles},
     ActionData, ActionKey, BoundPoseType, Input,
 };
-use crate::input::action_manifest_helpers::{
+use crate::input::action_manifest::helpers::{
```
Supreeeme, line 7:
```suggestion
use helpers::{
```

src/input/action_manifest.rs
```diff
@@ -203,12 +235,13 @@ struct DefaultBindings {
...
 }
 
-#[derive(Deserialize, Debug, PartialEq, Eq)]
+#[derive(Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
 #[serde(rename_all = "snake_case")]
 pub(super) enum ControllerType {
     ViveController,
     Knuckles,
     OculusTouch,
+    KhrSimpleController,
```
Supreeeme, line 244: I still don't like this - I don't understand why this is being added when no real game is going to do this.

knah: Alright, I dropped it completely, together with a commit related to it. Also rebased on latest main.

Supreeeme:
Okay, I think I'm satisfied with how this looks now. Thanks!

PR #63, opened 2025-03-14
Support overlay curvature

galister:
Some older games and mods use curved overlays, this makes them actually curved properly.

I know from my wlx-related research that this lines up exactly with what SteamVR would show. Looked good in testing as well.

Feedback is very welcome.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -11,13 +11,13 @@ use crate::{
 use log::{debug, info, trace};
 use openvr as vr;
 use openxr as xr;
-use std::ffi::c_char;
 use std::mem::offset_of;
 use std::sync::{
     atomic::{AtomicU32, Ordering},
     Arc, Mutex,
 };
 use std::time::Instant;
+use std::{ffi::c_char, ops::Deref};
 
 #[derive(Default)]
 pub struct CompositorSessionData(Mutex<Option<DynFrameController>>);
@@ -1028,7 +1028,7 @@ impl<G: GraphicsBackend> FrameController<G> {
         let overlay_layers;
         if let Some(overlay_man) = overlays {
             overlay_layers = overlay_man.get_layers(session_data);
-            layers.extend(overlay_layers.iter().map(std::ops::Deref::deref));
+            layers.extend(overlay_layers.iter().map(Deref::deref));
         }
 
         self.stream
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -3,13 +3,14 @@ use crate::{
     graphics_backends::{supported_apis_enum, GraphicsBackend, SupportedBackend},
     openxr_data::{GraphicalSession, OpenXrData, Session, SessionData},
 };
+use glam::{vec3, Quat, Vec3};
 use log::{debug, trace};
 use openvr as vr;
 use openxr as xr;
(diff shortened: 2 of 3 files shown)


src/compositor.rs
```diff
@@ -1028,7 +1028,10 @@ impl<G: GraphicsBackend> FrameController<G> {
         let overlay_layers;
         if let Some(overlay_man) = overlays {
             overlay_layers = overlay_man.get_layers(session_data);
-            layers.extend(overlay_layers.iter().map(std::ops::Deref::deref));
+            layers.extend(overlay_layers.iter().map(|l| match l {
+                OverlayLayer::Quad(quad) => quad.deref(),
+                OverlayLayer::Cylinder(cylinder) => cylinder.deref(),
+            }));
```
Supreeeme, line 1034: This could go in a Deref impl for `OverlayLayer`, thus removing the need to expose it here.

src/overlay.rs
```diff
@@ -68,48 +69,100 @@ impl OverlayMan {
...
-                // Internally, CompositionLayerQuad is using the raw OpenXR handles and PhantomData, not actual
-                // references, so returning it as long as we can guarantee the lifetimes of the space and
-                // swapchain is fine. Both of these are derived from the SessionData,
-                // so we should have no lifetime problems.
-                unsafe { xr::CompositionLayerQuad::from_raw(layer.into_raw()) }
-            }
+            // SAFETY: SetOverlayCurvature ensures the following:
+            // - Curvature is only ever non-zero if CompositionLayerCylinderKHR is supported.
+            // - Curvature is within [0.0, 1.0]
+            if overlay.curvature > f32::EPSILON {
```
Supreeeme, line 89: Why epsilon?

src/overlay.rs
```diff
@@ -68,48 +69,100 @@ impl OverlayMan {
...
+
+                let layer = xr::CompositionLayerCylinderKHR::new()
+                    .space(space)
+                    .layer_flags(xr::CompositionLayerFlags::BLEND_TEXTURE_SOURCE_ALPHA)
+                    .eye_visibility(xr::EyeVisibility::BOTH)
+                    .sub_image(
+                        xr::SwapchainSubImage::new()
+                            .image_array_index(vr::EVREye::Left as u32)
+                            .swapchain(swapchain)
+                            .image_rect(rect),
```
Supreeeme, line 111: This could be created before the if/else block since it's the same for both layers.

src/overlay.rs
```diff
@@ -68,48 +69,100 @@ impl OverlayMan {
...
+                    .pose(xr::Posef {
+                        orientation: pose.orientation,
+                        position: xr::Vector3f {
+                            x: center.x,
+                            y: center.y,
+                            z: center.z,
+                        },
+                    });
+
+                fn lifetime_extend<'a, 'b: 'a, G: xr::Graphics>(
```
Supreeeme, line 125: I hate to repeat this function, but the only other solutions would be a macro or trait... Not important, just want to vent.

galister:
applied the recommendations

src/overlay.rs
galister: I also have a 3rd type of layer now on my local, because I'm implementing the skybox by creating an internal-use overlay. so i will look into this in that next PR.

PR #64, opened 2025-03-15
IVRSettings stub

knah:
The most basic `IVRSettings` stub that returns zeroes and ignores writes while logging everything.
This upgrades Vivecraft from no-launch to playable (tested with WiVRn+q3).

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1,0 +1,178 @@
+use log::debug;
+use openvr as vr;
+use openvr::EVRSettingsError;
+use std::ffi::CStr;
+use std::os::raw::c_char;
+
+#[derive(Default, macros::InterfaceImpl)]
+#[interface = "IVRSettings"]
+#[versions(003)]
+pub struct Settings {
+    vtables: Vtables,
+}
+
+impl vr::IVRSettings003_Interface for Settings {
+    fn GetSettingsErrorNameFromEnum(&self, error: EVRSettingsError) -> *const c_char {
+        #[allow(unreachable_patterns)]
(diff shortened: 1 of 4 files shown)


src/settings.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn SetBool(
+        &self,
+        section: *const c_char,
+        settings_key: *const c_char,
+        value: bool,
+        error: *mut EVRSettingsError,
+    ) {
+        let section = unsafe { CStr::from_ptr(section) }.to_string_lossy();
+        let key = unsafe { CStr::from_ptr(settings_key) }.to_string_lossy();
+        info!("Setting bool on {section}/{key} to {value}");
```
Supreeeme, line 39: Would prefer if all of these logs were debug

src/settings.rs
```diff
@@ -0,0 +1,178 @@
...
+        let error: &'static str = match error {
+            EVRSettingsError::None => "",
+            EVRSettingsError::IPCFailed => "IPC Failed",
+            EVRSettingsError::WriteFailed => "Write Failed",
+            EVRSettingsError::ReadFailed => "Read Failed",
+            EVRSettingsError::JsonParseFailed => "JSON Parse Failed",
+            EVRSettingsError::UnsetSettingHasNoDefault => "Unset setting has no default",
+            EVRSettingsError::AccessDenied => "Access denied",
+            _ => "Unknown error",
+        };
```
Supreeeme, line 26: To be safe you should just make these `&'static CStr`. You can make a CStr literal like `c"IPC Failed"`.

PR #68, opened 2025-03-19
chore: Refactor device logic

### LINKED ISSUE #57: No controllers detected when only right controller is on
Will probably have to shift around the device index.

RinLovesYou:
This is the start of the device refactor work. I've decided to keep it as minimal as possible for now. This gives us enough of a skeleton to implement new devices that provide poses. Devices are now stored in a `Vec`, accessed via the `TrackedDeviceList` struct. They currently only provide a pose, and access to the device's interaction profile.

Passes all tests (once `Injected<T>::get()` is fixed).

Fixes #57 

~~Pulls in one new dependency: [enum_dispatch](https://docs.rs/enum_dispatch/latest/enum_dispatch/).
Storing a list of multiple types that implement the same trait can get expensive. Using an enum mostly solves this, but means we'd need to implement that trait for the enum. This crate automates this process for us.~~

### DIFF
--- a/src/input/devices.rs
+++ b/src/input/devices.rs
@@ -70,8 +70,8 @@ impl XrTrackedDevice {
 
         *pose_cache = match self.device_type {
             TrackedDeviceType::Hmd => self.get_hmd_pose(xr_data, session_data, origin),
-            TrackedDeviceType::Controller { .. } => {
-                self.get_controller_pose(xr_data, session_data, origin)
+            TrackedDeviceType::Controller { hand } => {
+                self.get_controller_pose(xr_data, session_data, hand, origin)
             }
         };
 
@@ -126,11 +126,12 @@ impl XrTrackedDevice {
         &self,
         xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
         session_data: &SessionData,
+        hand: Hand,
         origin: vr::ETrackingUniverseOrigin,
     ) -> Option<vr::TrackedDevicePose_t> {
         let legacy_actions = session_data.input_data.legacy_actions.get()?;
 
-        let spaces = match self.get_controller_hand()? {
+        let spaces = match hand {
             Hand::Left => &legacy_actions.left_spaces,
             Hand::Right => &legacy_actions.right_spaces,
         };


src/input/devices/tracked_device.rs
```diff
@@ -0,0 +1,177 @@
...
+    openxr_data::{AtomicPath, Hand, OpenXrData, SessionData},
+};
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[allow(dead_code)]
+pub enum TrackedDeviceType {
+    HMD,
+    LeftHand,
+    RightHand,
+    GenericTracker,
```
Supreeeme, line 23: I get that the point of this is for introducing the tracker functionality, but I'd prefer that all of the tracker related stuff be introduced together.

src/input/devices/tracked_device.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    fn get_interaction_profile(&self) -> Option<&'static dyn InteractionProfile> {
+        self.get_base_device()
+            .interaction_profile
+            .lock()
+            .unwrap()
+            .as_ref()
+            .copied()
+    }
+}
```
Supreeeme, line 136: I wonder, is exposing a trait for this valuable? Given that the majority of the methods just use some functionality from `BaseDevice`, would it not be simpler to just use the BaseDevice type directly? Otherwise it seems to be introducing an unnecessary abstraction (and dependency). You could expose the device specific functionality as part of the `TrackedDeviceType` enum, i.e.:
```rust
pub enum TrackedDeviceType {
    Hmd,
    Hand {
        hand: Hand,
        path: &'static str,
        subaction_path: xr::Path
    }
```

The only method that even appears to need per device type functionality is `get_pose`, and I can live with just matching on the device type inside of it.

src/input/devices/tracked_device.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+}
+
+impl BaseDevice {
+    pub fn new(device_index: vr::TrackedDeviceIndex_t, device_type: TrackedDeviceType) -> Self {
+        assert!(
+            device_index != vr::k_unTrackedDeviceIndexInvalid,
+            "Cannot create a device with an invalid index"
+        );
```
Supreeeme, line 165: The device itself being aware of its index seems pointless and error prone.

src/input/devices/tracked_device.rs
```diff
@@ -0,0 +1,177 @@
...
+};
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+#[allow(dead_code)]
+pub enum TrackedDeviceType {
+    HMD,
+    LeftHand,
+    RightHand,
+    GenericTracker,
+    Unknown,
```
Supreeeme, line 24: Not a fan of enum variants that are essentially invalid.

src/openxr_data.rs
```diff
@@ -142,32 +138,52 @@ impl<C: Compositor> OpenXrData<C> {
...
+                        continue;
+                    }
+
+                    let xr_input = self.input.get().unwrap();
+
+                    let devices = xr_input.devices.read().unwrap();
+
+                    for hand in [TrackedDeviceType::LeftHand, TrackedDeviceType::RightHand] {
+                        let controller = devices.get_controller(hand);
+                        let hmd = devices.get_hmd();
```
Supreeeme, line 151: This should move outside the loop

src/input/devices.rs
```diff
@@ -0,0 +1,127 @@
...
+// enum dispatch generates the trait implementation for this enum,
+// so we can access the methods of the variants without having to match anything manually.
+#[enum_dispatch(TrackedDevice)]
+pub enum TrackedDeviceContainer {
+    HMD(XrHMD),
+    Controller(XrController),
+}
+
+pub struct TrackedDeviceList {
+    pub devices: Vec<TrackedDeviceContainer>,
```
Supreeeme, line 27: Does this need to be pub?

src/input/devices/tracked_device.rs
RinLovesYou: I can see where you're coming from but device specific logic will increase with props for instance. Yes, we can structure it like this, and maybe move the device specific functions into different files so it's not all one huge file, we'll end up with `get_controller_string_tracked_property`, `get_hmd_string_tracked_property`, `get_generic_tracker_string_tracked_property`, etc... for every device type in the future. While with this it'd all just be `device.get_string_tracked_property()`, and it being a trait object would take care of matching them all for us.

src/openxr_data.rs
```diff
@@ -142,32 +138,52 @@ impl<C: Compositor> OpenXrData<C> {
...
                 xr::Event::InteractionProfileChanged(_) => {
                     let session = self.session_data.get();
-                    for info in [&self.left_hand, &self.right_hand] {
+                    if self.input.get().is_none() {
+                        continue;
+                    }
+
+                    let xr_input = self.input.get().unwrap();
+
+                    let devices = xr_input.devices.read().unwrap();
```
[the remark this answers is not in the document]
RinLovesYou: While that's true, I'd be creating multiple functions: `set_controller_subaction_path`, `set_device_interaction_profile`, `set_device_connected`, `get_controller_hand_path` (well, that last one can probably be a constant honestly), but the point is, all of these functions would be looping over the device list. In this alone it would amount to at least 10 times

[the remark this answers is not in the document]
Supreeeme: Why would you need to expose a bunch of methods? Just move all of this handling into `Input`.

src/input/devices/tracked_device.rs
Supreeeme: I don't see how adding several intermediary functions, which is really only happening for properties, is worse than adding several trait implementations.

src/openxr_data.rs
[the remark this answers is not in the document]
RinLovesYou: ah, you meant moving the entire logic to input. Fair enough, done

src/input/devices/controller.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+impl Default for ControllerVariables {
+    fn default() -> Self {
+        Self {
+            hand: Hand::Left,
+            subaction_path: xr::Path::default(),
+        }
+    }
+}
```
Supreeeme, line 23: I don't see why this has a Default impl.

src/input/devices/hmd.rs
```diff
@@ -0,0 +1,26 @@
+use openvr as vr;
+
```
Supreeeme, line 2: This file, controller.rs and tracked_device.rs are such small files I think they can all just go into devices.rs.

src/input/devices/controller.rs
```diff
@@ -0,0 +1,76 @@
...
+            subaction_path: match hand {
+                Hand::Left => instance.string_to_path(hand.into()).unwrap(),
+                Hand::Right => instance.string_to_path(hand.into()).unwrap(),
+            },
+        }
+    }
+}
+
+impl XrTrackedDevice {
+    pub fn get_controller_pose(
```
Supreeeme, line 38: This does not need to be `pub`, should be `pub(super)` (or not pub at all if it goes in devices.rs

src/input/devices/hmd.rs
```diff
@@ -0,0 +1,26 @@
...
+use crate::openxr_data::{OpenXrData, SessionData};
+
+use super::tracked_device::XrTrackedDevice;
+
+impl XrTrackedDevice {
+    pub fn get_hmd_pose(
```
Supreeeme, line 8: `pub(super)`

src/input/action_manifest.rs
```diff
@@ -77,14 +77,24 @@ impl<C: openxr_data::Compositor> Input<C> {
...
 
+        let devices = self.devices.read().unwrap();
+        let left_hand = devices
+            .get_controller(Hand::Left)
+            .get_controller_variables()
+            .ok_or(vr::EVRInputError::InvalidDevice)?;
+        let right_hand = devices
+            .get_controller(Hand::Right)
+            .get_controller_variables()
+            .ok_or(vr::EVRInputError::InvalidDevice)?;
```
Supreeeme, line 88: This would require controllers to be connected to successfully load an action manifest, which is not ideal.

src/input/devices.rs
```diff
@@ -0,0 +1,78 @@
...
+            devices: vec![
+                XrTrackedDevice::new(TrackedDeviceType::Hmd),
+                XrTrackedDevice::new(TrackedDeviceType::Controller(ControllerVariables::new(
+                    xr_instance,
+                    Hand::Left,
+                ))),
+                XrTrackedDevice::new(TrackedDeviceType::Controller(ControllerVariables::new(
+                    xr_instance,
+                    Hand::Right,
+                ))),
```
Supreeeme, line 30: Controllers should not be created if they don't actually exist.

src/input/devices.rs
```diff
@@ -0,0 +1,78 @@
...
+        unsafe { self.devices.get_unchecked(hand as usize) }
+    }
+
+    pub fn iter(&self) -> std::slice::Iter<'_, XrTrackedDevice> {
+        self.devices.iter()
+    }
+
+    pub fn len(&self) -> usize {
+        self.devices.len()
+    }
```
Supreeeme, line 77: Is this used anywhere?

src/input/devices/controller.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use super::tracked_device::{TrackedDeviceType, XrTrackedDevice};
+
+use log::trace;
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct ControllerVariables {
+    pub hand: Hand,
+    pub subaction_path: xr::Path,
+}
```
Supreeeme, line 14: I'm not sure if there's a particular benefit to this living in a separate struct.

src/input/devices/hmd.rs
RinLovesYou: I think that will get a little messy if we add more things like props, and more devices like trackers. Personally i think It'll be easier to overview in separate files

src/input/action_manifest.rs
RinLovesYou: no it does not, they do not need to be connected for these variables to exist

src/input/devices/tracked_device.rs
```diff
@@ -0,0 +1,106 @@
...
+        self.connected.load(Ordering::Relaxed)
+    }
+
+    pub fn set_connected(&self, connected: bool) {
+        self.connected.store(connected, Ordering::Relaxed);
+    }
+
+    pub fn get_type(&self) -> TrackedDeviceType {
+        self.device_type
+    }
```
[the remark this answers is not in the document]
RinLovesYou: This question conflicts with the above one

src/input/action_manifest.rs
RinLovesYou: Do you want me to go back to storing the subaction path in a separate struct?

src/input/devices.rs
```diff
@@ -0,0 +1,73 @@
...
+
+mod controller;
+mod hmd;
+pub mod tracked_device;
+
+pub struct TrackedDeviceList {
+    devices: Vec<XrTrackedDevice>,
+}
+
+#[allow(dead_code)]
```
Supreeeme, line 16:
```suggestion
```

src/input/devices/tracked_device.rs
```diff
@@ -0,0 +1,117 @@
...
+        session_data: &SessionData,
+        origin: vr::ETrackingUniverseOrigin,
+    ) -> Option<vr::TrackedDevicePose_t> {
+        match self.device_type {
+            TrackedDeviceType::Hmd => self.get_hmd_pose(xr_data, session_data, origin),
+            TrackedDeviceType::Controller { .. } => {
+                self.get_controller_pose(xr_data, session_data, origin)
+            }
+        }
+    }
```
Supreeeme, line 77: Poses are no longer cached it seems.

src/input/devices/tracked_device.rs
```diff
@@ -0,0 +1,117 @@
...
+        match value {
+            0 => Ok(Self::Hmd),
+            1 => Ok(Self::Controller {
+                hand: Hand::Left,
+                subaction_path: xr::Path::default(),
+            }),
+            2 => Ok(Self::Controller {
+                hand: Hand::Right,
+                subaction_path: xr::Path::default(),
+            }),
```
Supreeeme, line 36: This is problematic - see #57. I also dislike the way this is used - creating a struct just to get the type feels bad.

src/input/devices/tracked_device.rs
RinLovesYou: Yeah i wasn't sure how to best bring this back, should i have every device hold onto its own pose cache?

src/input/devices/hmd.rs
Supreeeme: When the file actually becomes unwieldy to read through. As it stands I don't see this being any worse than input.rs or action_manifest.rs is right now.

src/input/devices/tracked_device.rs
Supreeeme: That does sound like the simplest approach.

src/input/devices/tracked_device.rs
RinLovesYou: Yeah it looks to me like I'll have to solve #57 anyways to only create controllers when they exist, I'll get on it

src/input/action_manifest.rs
RinLovesYou: doneso

src/input/devices/tracked_device.rs
RinLovesYou: done

src/input.rs
```diff
@@ -47,12 +51,13 @@ pub struct Input<C: openxr_data::Compositor> {
     action_map: RwLock<SlotMap<ActionKey, Action>>,
     set_map: RwLock<SlotMap<ActionSetKey, String>>,
     loaded_actions_path: OnceLock<PathBuf>,
-    cached_poses: Mutex<CachedSpaces>,
     legacy_state: legacy::LegacyState,
     skeletal_tracking_level: RwLock<vr::EVRSkeletalTrackingLevel>,
     profile_map: HashMap<xr::Path, &'static profiles::ProfileProperties>,
     estimated_finger_state: [Mutex<FingerState>; 2],
+    subaction_paths: [Mutex<xr::Path>; 2],
```
Supreeeme, line 58: I think this should just be a struct with `left` and `right` members, for clarity's sake. Some function like `to_iter` can be added if iteration is important.

src/input.rs
```diff
@@ -107,34 +113,47 @@ impl<C: openxr_data::Compositor> Input<C> {
...
             estimated_finger_state: [
                 Mutex::new(FingerState::new()),
                 Mutex::new(FingerState::new()),
             ],
+            subaction_paths,
             events: Mutex::default(),
         }
     }
 
+    pub fn get_subaction_path(&self, hand: Hand) -> xr::Path {
```
Supreeeme, line 143: Does not need to be pub as far as I can tell.

src/input/devices.rs
RinLovesYou: This was a big one, should be working now.

src/input/devices.rs
```diff
@@ -0,0 +1,411 @@
...
+}
+
+pub struct TrackedDeviceCreateInfo {
+    pub device_type: TrackedDeviceType,
+    pub profile_path: Option<xr::Path>,
+    pub interaction_profile: Option<&'static dyn InteractionProfile>,
+}
+
+impl XrTrackedDevice {
+    pub fn new(info: TrackedDeviceCreateInfo) -> Self {
```
Supreeeme, line 39:
```suggestion
    pub(super) fn new(info: TrackedDeviceCreateInfo) -> Self {
```

src/input/devices.rs
```diff
@@ -0,0 +1,411 @@
...
+    connected: AtomicBool,
+    previous_connected: AtomicBool,
+    pose_cache: Mutex<Option<vr::TrackedDevicePose_t>>,
+}
+
+pub struct TrackedDeviceCreateInfo {
+    pub device_type: TrackedDeviceType,
+    pub profile_path: Option<xr::Path>,
+    pub interaction_profile: Option<&'static dyn InteractionProfile>,
+}
```
Supreeeme, line 36: Why does this need to be a struct?

src/input/devices.rs
```diff
@@ -0,0 +1,411 @@
...
+        }
+    }
+
+    pub fn get_pose(
+        &self,
+        xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
+        session_data: &SessionData,
+        origin: vr::ETrackingUniverseOrigin,
+    ) -> Option<vr::TrackedDevicePose_t> {
+        let mut pose_cache = self.pose_cache.lock().ok()?;
```
Supreeeme, line 66:
```suggestion
        let mut pose_cache = self.pose_cache.lock().unwrap();
```
Mutex poisoning is not something I intend to ever try and recover from.

src/input/devices.rs
```diff
@@ -0,0 +1,411 @@
...
+        xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
+        session_data: &SessionData,
+        origin: vr::ETrackingUniverseOrigin,
+    ) -> Option<vr::TrackedDevicePose_t> {
+        let legacy_actions = session_data.input_data.legacy_actions.get()?;
+
+        let spaces = match self.get_controller_hand()? {
+            Hand::Left => &legacy_actions.left_spaces,
+            Hand::Right => &legacy_actions.right_spaces,
+        };
```
Supreeeme, line 136: This function should just require the Hand to be passed in.

src/input.rs
```diff
@@ -4,22 +4,29 @@ mod legacy;
...
+
 #[cfg(test)]
 mod tests;
 
-use profiles::MainAxisType;
+use devices::SubactionPaths;
+use devices::TrackedDeviceCreateInfo;
+use devices::TrackedDeviceList;
+use devices::TrackedDeviceType;
+use devices::XrTrackedDevice;
```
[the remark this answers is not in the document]
RinLovesYou: yeah normally this is taken care of automatically by rust-analyzer, weird it didn't do this here.

src/input.rs
```diff
@@ -4,22 +4,29 @@ mod legacy;
 mod profiles;
 mod skeletal;
 
+pub mod devices;
+
```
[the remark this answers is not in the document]
RinLovesYou: this actually needs to be a `pub use devices::TrackedDeviceType`

src/input/devices.rs
RinLovesYou: basically, since rust is very particular about mutable/immutable access of the device list, i need to push device creation to the end of the `interaction_profile_changed` function, so i store these in a struct which seems like the most straight forward approach. See `src/input.rs:1239`

src/input/devices.rs
RinLovesYou: can you elaborate?

Supreeeme: Instead of calling `get_controller_hand`, just pass the Hand in as an argument.

src/input.rs
```diff
@@ -1146,71 +1172,97 @@ impl<C: openxr_data::Compositor> vr::IVRInput005On006 for Input<C> {
...
+                if let Some(controller) = controller {
+                    controller.set_interaction_profile(p);
+                } else {
+                    devices_to_create.push(TrackedDeviceCreateInfo {
+                        device_type: TrackedDeviceType::Controller { hand },
+                        profile_path: Some(profile_path),
+                        interaction_profile: Some(p),
+                    });
+                }
+                hmd.set_interaction_profile(p);
```
Supreeeme, line 1222: I'm not sure the hmd having an interaction profile is particularly meaningful

src/input.rs
```diff
@@ -1337,16 +1328,41 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+            if device.compare_exchange_connected().is_ok() {
+                debug!(
+                    "sending {:?} {}connected",
+                    device.get_type(),
+                    if current { "" } else { "not " }
+                );
+
+                // VREvent_t can be different sizes depending on the OpenVR version,
+                // so we use raw pointers to avoid creating a reference, because if the
+                // size doesn't match our VREvent_t's size, we are in UB land
```
Supreeeme, line 1345: This comment is in the wrong place, it should stay where it originally was.

src/input/devices.rs
Supreeeme: Given that this is only used in `interaction_profile_changed`, it can just be moved inside of that function, and `XrTrackedDevice::new` can just take the parameters.

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub fn compare_exchange_connected(&self) -> Result<bool, bool> {
+        let current = self.connected();
+
+        self.previous_connected.compare_exchange(
+            !current,
+            current,
+            Ordering::Relaxed,
+            Ordering::Relaxed,
+        )
+    }
```
Supreeeme, line 118: The way this function is used, it always just checks `is_ok`, so returning the Result seems kind of pointless.

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+            return Err(vr::EVRInputError::MaxCapacityReached);
+        }
+
+        self.devices.push(device);
+
+        Ok(index)
+    }
+
+    pub(super) fn get_hmd(&self) -> &XrTrackedDevice {
+        unsafe { self.devices.get_unchecked(0) }
```
Supreeeme, line 241:
```suggestion
        self.devices.get(0).unwrap()
```
I get this should never fail, but no point in going unsafe.

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub fn get_poses(
+        &self,
+        poses: &mut [vr::TrackedDevicePose_t],
+        origin: Option<vr::ETrackingUniverseOrigin>,
+    ) {
+        tracy_span!();
+        let devices = self.devices.read().unwrap();
+        let session_data = self.openxr.session_data.get();
+
+        poses.iter_mut().enumerate().for_each(|(i, pose)| {
```
Supreeeme, line 271:
```suggestion
       for (i, pose) in poses.iter_mut().enumerate()
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+            }
+        });
+    }
+
+    pub fn get_controller_pose(
+        &self,
+        hand: Hand,
+        origin: Option<vr::ETrackingUniverseOrigin>,
+    ) -> Option<vr::TrackedDevicePose_t> {
+        let controller_index = self.devices.read().ok()?.get_controller_index(hand);
```
Supreeeme, line 291:
```suggestion
        let controller_index = self.devices.read().unwrap().get_controller_index(hand);
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub fn get_device_pose(
+        &self,
+        index: vr::TrackedDeviceIndex_t,
+        origin: Option<vr::ETrackingUniverseOrigin>,
+    ) -> Option<vr::TrackedDevicePose_t> {
+        tracy_span!();
+
+        let session_data = self.openxr.session_data.get();
+
+        self.devices.read().ok()?.get_device(index)?.get_pose(
```
Supreeeme, line 305:
```suggestion
        self.devices.read().unwrap().get_device(index)?.get_pose(
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+            &self.openxr,
+            &session_data,
+            origin.unwrap_or(session_data.current_origin),
+        )
+    }
+
+    pub fn is_device_connected(&self, index: vr::TrackedDeviceIndex_t) -> bool {
+        let Some(devices) = self.devices.read().ok() else {
+            return false;
+        };
```
Supreeeme, line 315:
```suggestion
    let devices = self.devices.read().unwrap();
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        let Some(device) = devices.get_device(index) else {
+            return false;
+        };
+
+        device.connected()
```
Supreeeme, line 321:
```suggestion
        devices.get_device(index).is_some_and(|device| device.connected());
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    pub fn device_index_to_device_type(
+        &self,
+        index: vr::TrackedDeviceIndex_t,
+    ) -> Option<TrackedDeviceType> {
+        let devices = self.devices.read().ok()?;
```
Supreeeme, line 328:
```suggestion
        let devices = self.devices.read().unwrap();
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+        let device = devices.get_device(index)?;
+
+        Some(device.get_type())
+    }
+
+    pub fn device_index_to_hand(&self, index: vr::TrackedDeviceIndex_t) -> Option<Hand> {
+        let devices = self.devices.read().ok()?;
```
Supreeeme, line 335:
```suggestion
        let devices = self.devices.read().unwrap();
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+        let device = devices.get_device(index)?;
+
+        device.get_controller_hand()
+    }
+
+    pub fn get_controller_device_index(&self, hand: Hand) -> Option<vr::TrackedDeviceIndex_t> {
+        let devices = self.devices.read().ok()?;
```
Supreeeme, line 342:
```suggestion
        let devices = self.devices.read().unwrap();
```

src/input/devices.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        if controller_index == vr::k_unTrackedDeviceIndexInvalid {
+            return None;
+        }
+
+        Some(controller_index)
+    }
+
+    fn get_profile_data(&self, hand: Hand) -> Option<&super::profiles::ProfileProperties> {
+        let devices = self.devices.read().ok()?;
```
Supreeeme, line 353:
```suggestion
        let devices = self.devices.read().unwrap();
```

src/system.rs
```diff
@@ -258,13 +247,21 @@ impl vr::IVRSystem022_Interface for System {
         state_size: u32,
         pose: *mut vr::TrackedDevicePose_t,
     ) -> bool {
+        let Some(input) = self.input.get() else {
+            return false;
+        };
```
Supreeeme, line 252:
```suggestion
        let input = self.input.force(|_| Input::new(self.openxr.clone()));
```

src/system.rs
```diff
@@ -546,12 +505,22 @@ impl vr::IVRSystem022_Interface for System {
...
             *err = vr::ETrackedPropertyError::Success;
         }
         match device_index {
-            x if Hand::try_from(x).is_ok() => self.input.get().and_then(|input| {
-                input.get_controller_int_tracked_property(Hand::try_from(x).unwrap(), prop)
+            x if input.device_index_to_hand(x).is_some() => self.input.get().and_then(|input| {
+                input.get_controller_int_tracked_property(
+                    input.device_index_to_hand(x).unwrap(),
+                    prop,
+                )
```
Supreeeme, line 523:
```suggestion
            x if input.device_index_to_hand(x).is_some() => input.get_controller_int_tracked_property(
                    input.device_index_to_hand(x).unwrap(),
                    prop,
                )
```

src/system.rs
```diff
@@ -471,8 +427,11 @@ impl vr::IVRSystem022_Interface for System {
                 | vr::ETrackedDeviceProperty::ControllerType_String => Some(c"<unknown>"),
                 _ => None,
             },
-            x if Hand::try_from(x).is_ok() => self.input.get().and_then(|i| {
-                i.get_controller_string_tracked_property(Hand::try_from(x).unwrap(), prop)
+            x if input.device_index_to_hand(x).is_some() => self.input.get().and_then(|i| {
+                i.get_controller_string_tracked_property(
+                    input.device_index_to_hand(x).unwrap(),
+                    prop,
+                )
```
Supreeeme, line 434:
```suggestion
            x if input.device_index_to_hand(x).is_some() =>
                input.get_controller_string_tracked_property(
                    input.device_index_to_hand(x).unwrap(),
                    prop,
                )
```

src/system.rs
```diff
@@ -604,62 +573,49 @@ impl vr::IVRSystem022_Interface for System {
...
-            vr::k_unTrackedDeviceIndex_Hmd => true,
-            x if Hand::try_from(x).is_ok() => match Hand::try_from(x).unwrap() {
-                Hand::Left => self.openxr.left_hand.connected(),
-                Hand::Right => self.openxr.right_hand.connected(),
-            },
-            _ => false,
-        }
+        self.input
+            .get()
+            .is_some_and(|input| input.is_device_connected(device_index))
```
Supreeeme, line 578: This is relying on input to be set up for the hmd to be marked as connected.

src/system.rs
```diff
@@ -604,62 +573,49 @@ impl vr::IVRSystem022_Interface for System {
...
-                if self.IsTrackedDeviceConnected(x) {
-                    vr::ETrackedDeviceClass::Controller
-                } else {
-                    vr::ETrackedDeviceClass::Invalid
+        self.input
+            .get()
+            .and_then(|input| match input.device_index_to_device_type(index) {
+                Some(TrackedDeviceType::Hmd) => Some(vr::ETrackedDeviceClass::HMD),
+                Some(TrackedDeviceType::Controller { .. }) => {
+                    Some(vr::ETrackedDeviceClass::Controller)
```
Supreeeme, line 587: Same comment as previous

src/system.rs
```diff
@@ -673,9 +629,13 @@ impl vr::IVRSystem022_Interface for System {
...
         device_index: vr::TrackedDeviceIndex_t,
     ) -> vr::EDeviceActivityLevel {
+        let Some(input) = self.input.get() else {
+            return vr::EDeviceActivityLevel::Unknown;
+        };
+
         match device_index {
             vr::k_unTrackedDeviceIndex_Hmd => vr::EDeviceActivityLevel::UserInteraction,
-            x if Hand::try_from(x).is_ok() => {
+            x if input.device_index_to_hand(x).is_some() => {
```
Supreeeme, line 638: Same comment as previous

Supreeeme (commented):
Looks like this could use a rebase.

All the unwrap suggestions are because they are on a RwLock, and I'm not going to handle RwLock poisoning.

src/input.rs
TayouVR: HMDs can have buttons, like the one on the vive opening the SteamVR dashboard by default. So it would make sense, wouldn't it?

Supreeeme: There is no way of utilizing the HMD buttons in OpenXR, and this code is setting the HMD profile to match the controllers.

Beyley: > There is no way of utilizing the HMD buttons in OpenXR, and this code is setting the HMD profile to match the controllers.

This isn't necessarily the case, see https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#_htc_vive_pro_profile

Some HMDs have interaction profiles. But yes, as is this code is setting it to the same as the controllers, so that isn't terribly relevant here.

Supreeeme:
Superseded by #223

PR #69, opened 2025-03-19
Implement Skeletal Summary

RinLovesYou:
ports over the skeletal summary from OpenComposite to make gestures work in VRChat on index controllers. I'd like to do this better in the future (and implement splay), but this at least gets things working.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -491,17 +491,21 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
     fn GetSkeletalSummaryData(
         &self,
         action: vr::VRActionHandle_t,
-        _: vr::EVRSummaryType,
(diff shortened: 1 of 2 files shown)


Supreeeme:
Instead of having separate paths for estimated and real skeletal summaries, I think we can just
1. cache the skeleton pose each frame
2. use the same path for estimated and real skeletons

RinLovesYou:
> 1. cache the skeleton pose each frame
> 2. use the same path for estimated and real skeletons

As in, cache the openvr skeleton each frame and do a unified summary calculation from *that*?

Supreeeme:
Yes, but don't automatically get the skeletal state each frame. Just when necessary, and don't try to calculate it more than once a frame.

palodequeso:
I think this might fix https://github.com/Supreeeme/xrizer/issues/151, at least for me?

Issue #43, opened 2025-02-04
Derail Valley - no video in headset


AdamK2003:
Retested Derail Valley after #17 got fixed, the game is launching and the preview is showing up fine in the game window, but the headset displays a black screen.

Supreeeme:
Can you run the game with `RUST_LOG=debug` and provide the log?

AdamK2003:
Log here: 

[xrizer.txt.gz](https://github.com/user-attachments/files/18664069/xrizer.txt.gz)

Supreeeme:
Strange. Can you provide another log with `RUST_LOG=trace`

AdamK2003:
New log here:

[xrizer.txt.gz](https://github.com/user-attachments/files/18776852/xrizer.txt.gz)

Supreeeme:
This likely has changed behavior wise since cd87a81df98e3efd6262804282aa5d22735bd44f, can you test again and provide a log if still not working?

AdamK2003:
My right controller is busted right now, which makes VR Kinda Hard, but will check once the new one arrives

AdamK2003:
Update: finally tested, it's rendering fine now (in game controls aren't working, but that's a separate issue). Closing this.

PR #72, opened 2025-03-22
compositor: recreate the swapchain regardless of should_render

### LINKED ISSUE #36: Resonite doesn't show anything in the headset
With 50565f490e08d2cec58cdbb0e821f18ea0bd7ce1 reverted Resonite still doesn't render anything in the headset, even though it can be seen tracking the head pose in the desktop window.

With RUST_LOG=trace: [xrizer.txt.gz](https://github.com/user-attachments/files/18510261/xrizer.txt.gz)

It appears 680833891ec7b849d06879063578fc7b7d2dd982 is implicated here, reverting that and applying the following diff makes everything work again: (image_acquired check added because of a panic there)
```diff
diff --git a/src/compositor.rs b/src/compositor.rs
index 8d1d4a3..3706599 100644
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -99,7 +99,9 @@ impl Compositor {
             AnyTempBackendData: From<TempBackendData<G>>,
         {
             let b_texture = G::get_texture(texture);
-            let info = backend.swapchain_info_for_texture(b_texture, bounds, texture.eColorSpace);
+            let mut info = backend.swapchain_info_for_texture(b_texture, bounds, texture.eColorSpace);
+            info.width = info.width.max(1);
+            info.height = info.height.max(1);
             TempBackendData {
                 backend,
                 swapchain_create_info: info,
@@ -862,8 +864,10 @@ impl<G: GraphicsBackend> FrameController<G> {
         // Both eyes submitted: show our images
 
         trace!("releasing image");
-        self.swapchain_data.swapchain.release_image().unwrap();
-        self.image_acquired = false;
+        if self.image_acquired {
+            self.swapchain_data.swapchain.release_image().unwrap();
+            self.image_acquired = false;
+        }
 
         let mut proj_layer_views = Vec::new();
         if self.should_render {
```

notpeelz:
It seems like certain games (like Resonite) will first submit an empty texture before sending the real frames, which results in the game never rendering to the HMD.

This is what I think is happening:
- the game submits a frame of dimensions 0x0
- the frame controller is initialized using invalid swapchain dimensions
- the swapchain creation is skipped due to validation checks
- the frame controller calls `xrWaitFrame()`, which always returns `should_render = false` (that's because the session never moves past `XR_SESSION_STATE_READY`, but I don't know exactly why)
- game submits more frames, but because `should_render` is never `true` the swapchain is never re-created

Fixes #36 and possibly #43 (waiting for confirmation)

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -964,51 +964,49 @@ impl<G: GraphicsBackend> FrameController<G> {
             return Err(vr::EVRCompositorError::AlreadySubmitted);
         }
 
-        self.eyes_submitted[eye as usize] = if self.should_render {
-            // Make sure our image dimensions haven't changed.
-            let new_info = self
-                .backend
-                .swapchain_info_for_texture(texture, bounds, color_space);
+        // Make sure our image dimensions haven't changed.
+        let new_info = self
+            .backend
+            .swapchain_info_for_texture(texture, bounds, color_space);
 
-            is_valid_swapchain_info(&new_info)
-                .then(|| {
-                    assert!(
-                        !self.submitting_null,
-                        "App submitted a null texture and a normal texture in the same frame"
-                    );
-
-                    let creation_format;
-                    let current_info = if let Some(data) = self.swapchain_data.as_ref() {
-                        creation_format = data.initial_format;
-                        &data.info
-                    } else {
-                        // SAFETY: Technically SessionCreateInfo should be Copy anyway so this should be fine:
-                        // https://github.com/Ralith/openxrs/issues/183
-                        creation_format = new_info.format;
-                        self.recreate_swapchain(session_data, unsafe { std::ptr::read(&new_info) });
-                        self.swapchain_data.as_ref().map(|d| &d.info).unwrap()
-                    };
-                    if !is_usable_swapchain(current_info, creation_format, &new_info) {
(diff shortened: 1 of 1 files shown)


Supreeeme:
I figured out what's happening. The explanation you provide doesn't make sense because the session state should not be dependent on the swapchain state. I confirmed this by checking out the Monado code - the state after `READY` is `SYNCHRONIZED`, and it only ever changes to `SYNCHRONIZED` inside `xrEndFrame` - which leads to the main issue: `xrEndFrame` was not being called if the swapchain is invalid, due to this:
```rust
        let Some(swapchain_data) = self.swapchain_data.as_mut() else {
            return Err(vr::EVRCompositorError::None);
        };
```
When the initial swapchain info is invalid, we will never get past this point in `submit_impl`, even if we get a valid one later, because as you pointed out `should_render` is never true since we never get out of `READY`, so we will never have a valid swapchain info. Instead of unnecessarily recreating the swapchain info though, I'd rather just handle null submits properly and ensure `xrEndFrame` always gets called. This should be rather easy to do, so I'll just do it myself. I appreciate you looking into it though!

notpeelz:
Ah that makes a lot of sense. I knew there was something going on with the session state transitions, but couldn't quite figure out how it works (being new to OpenXR).

PR #76, opened 2025-03-23
compositor: refactor swapchain re-creation check

notpeelz:
Minor refactor to avoid the `unsafe { std::ptr::read(&new_info) }` workaround.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -977,16 +977,13 @@ impl<G: GraphicsBackend> FrameController<G> {
                         "App submitted a null texture and a normal texture in the same frame"
                     );
 
-                    if self.swapchain_data.as_ref().is_none() {
-                        // SAFETY: Technically SessionCreateInfo should be Copy anyway so this should be fine:
-                        // https://github.com/Ralith/openxrs/issues/183
-                        self.recreate_swapchain(session_data, unsafe { std::ptr::read(&new_info) });
-                    }
-                    let data = self.swapchain_data.as_ref().unwrap();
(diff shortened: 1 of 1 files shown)


src/compositor.rs
```diff
@@ -977,16 +977,22 @@ impl<G: GraphicsBackend> FrameController<G> {
...
+                        .as_ref()
+                        .map(|data| {
+                            !is_usable_swapchain(&data.info, data.initial_format, &new_info)
+                        })
+                        .inspect(|not_usable| {
+                            if *not_usable {
+                                info!("recreating swapchain (for {eye:?})");
+                            }
+                        })
+                        .unwrap_or(true)
```
Supreeeme, line 991: I think it's clearer if you reverse the condition:
```suggestion
                    if !self
                        .swapchain_data
                        .as_ref()
                        .is_some_and(|data| {
                            is_usable_swapchain(&data.info, data.initial_format, &new_info)
                        })
                        .inspect(|usable| {
                            if !usable {
                                info!("recreating swapchain (for {eye:?})");
                            }
                        })
                        .unwrap_or(false)
```

notpeelz: `is_some_and` returns a bool, so that won't work with `.inspect`.

notpeelz: What about a simple `match`? I find that a lot more readable, even if it means having two branches that call `recreate_swapchain()`.

Supreeeme: I don't see how a match would be significantly more readable. I'd just move the recreating swapchain log inside of the if block, I don't think it's that important to super precise with the log.

PR #78, opened 2025-03-26
implement overlay alpha using KHR_composition_layer_color_scale_bias

galister:
I refactored OverlayLayer into CompositorLayer and moved it into compositor.rs

adding to next chains in openxrs is always a pain, I tried to make it as isolated as possible

KHR_composition_layer_color_scale_bias is implemented by monado, but not enabled by default. we can have envision/wivrn enable it by default if it comes to that. i just tested by enabling it on wivrn.

also fixes the bounds using wrong dimensions

this gives a perfect loading screen in vrc, unless looking from a weird angle.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -1354,7 +1354,6 @@ mod tests {
             _texture: Self::OpenVrTexture,
             _bounds: openvr::VRTextureBounds_t,
             _image_index: usize,
-            _alpha: f32,
         ) -> openxr::Extent2Di {
             xr::Extent2Di::default()
         }
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -298,7 +298,7 @@ pub struct OverlayLayer<'a, G: xr::Graphics> {
     color_bias_khr: Option<Box<xr::sys::CompositionLayerColorScaleBiasKHR>>,
 }
 
-impl<'a, G: xr::Graphics> OverlayLayer<'a, G> {
+impl<G: xr::Graphics> OverlayLayer<'_, G> {
     pub fn set_alpha(&mut self, alpha: f32) {
         // only one instance is stored, so this would cause segfault due to UAF
         debug_assert!(
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -41,7 +41,7 @@ impl OverlayMan {
     pub fn set_skybox(&self, session: &SessionData, textures: &[vr::Texture_t]) {
         // We don't yet follow HMD position, so the skybox needs to be
         // big enough so that the user never leaves it
-        const SKYBOX_SIZE: f32 = 50.0;
+        const SKYBOX_SIZE: f32 = 500.0;
 
         self.clear_skybox();
 
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -8,10 +8,10 @@ use log::{debug, trace};
 use openvr as vr;
 use openxr as xr;
 use slotmap::{new_key_type, Key, KeyData, SecondaryMap, SlotMap};
-use std::collections::HashMap;
 use std::f32::consts::{FRAC_1_SQRT_2, PI};
 use std::ffi::{c_char, c_void, CStr, CString};
 use std::sync::{Arc, Mutex, RwLock};
+use std::{collections::HashMap, ops::Deref};
 
 // OpenVR overlays are allowed to use ≥ 0
 pub const SKYBOX_Z_ORDER: i64 = -1;
@@ -223,7 +223,9 @@ impl OverlayMan {
                         });
 
                     let layer = lifetime_extend!(CompositionLayerQuad, layer);
-                    layers.push((overlay.z_order, OverlayLayer::Quad(layer)));
+                    let mut layer = OverlayLayer::from(OverlayLayerInner::Quad(layer));
+                    overlay.alpha.iter().for_each(|a| layer.set_alpha(*a));
+                    layers.push((overlay.z_order, layer));
                 }
                 // SetOverlayCurvature checks for khr_composition_layer_cylinder
                 OverlayKind::Curved { curvature } => {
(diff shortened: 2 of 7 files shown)


Supreeeme (commented):
This overlay stuff desperately needs tests, but given I haven't figured out what those should look like yet I won't block this PR on that.

src/compositor.rs
```diff
@@ -1116,6 +1116,94 @@ fn is_valid_swapchain_info<G: xr::Graphics>(info: &xr::SwapchainCreateInfo<G>) -
...
+impl<'a, G: xr::Graphics> CompositorLayer<'a, G> {
+    pub fn as_base_layer(&self) -> &xr::CompositionLayerBase<'a, G> {
+        self.layer.deref()
+    }
+
+    pub fn set_alpha(&mut self, alpha: f32) {
+        debug_assert!(
+            self.color_bias_khr.is_none(),
+            "attempted to set_alpha on the same CompositorLayer twice!"
+        );
```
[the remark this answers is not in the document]
galister: this checks against calling set_alpha more than once on the same `OverlayLayer` within the same `get_layers` call. 

this is never intended to happen, however if it does, it would be a pain to debug. it would cause a segfault due to the unsafe code involved.

myghi63:
This is a huge improvement!
Current implementation:
Screenshot_org meumeu wivrn github testing_2025 03 27-10 09 51 572_232

New implementation, from this pull request:
Screenshot_org meumeu wivrn github testing_2025 03 27-14 38 55 576_075

Screenshot of the same game, but running on Pico 4 as standalone just for reference:
Screenshot_com vrchat android_2025 03 27-14 42 40 339_411

I've tested using WiVRn (master branch), with a AMD GPU.

If this pull request gets merged, then it will only have two differences:
-> The skybox of the loading screen looks like a cube, while on SteamVR and standalone game looks like a cilinder;
-> The picture of the world looks washed out.

Is the picture color issue related to KHR_composition_layer_color_scale_bias, which Galister said it's not enabled by default?

galister:
if you're using envision, add the cmake flag `XRT_FEATURE_OPENXR_LAYER_COLOR_SCALE_BIAS=ON` to your profile and then clean build

galister:
does it look less washed out with the alpha option compiled?

i can't think of another reason why it would be washed out. the loading message and skybox have the correct colors and all. it could be that the thumbnail is using an _SRGB format that dxvk translates as linear?

myghi63:
> does it look less washed out with the alpha option compiled?

Yep... but something is off. I've noticed that some worlds look better:
Screenshot_org meumeu wivrn github testing_2025 03 28-18 16 35 675_660

But, for whathever reason, the example I gave before looks the same:
Screenshot_org meumeu wivrn github testing_2025 03 28-18 17 23 257_325

galister:
added a commit which makes the skybox bigger, should make it feel less boxy

galister:
from my testing, the ones that are taken in unity seem ok, but the manually uploaded ones (ones that have logos/etc) have the gamma off?

myghi63:
I've done further testing and I was wrong, all pictures has the same problem and I deleted the last comment to avoid confusion.
The world I thought that looks alright is actually much darker on standalone.

And yes, about the skybox, it still looks like a cube, but it looks closer to what I see on standalone (thank you!!).
Is there another known game that also uses it?

galister:
no idea, most games would use an equirect (inside-out sphere), but vrchat uses a box sky of 6 square sides. there's no such thing as cylinder in steamvr

it's possible that they use cylinder or equirect on standalone

galister:
```
set overlay R8G8B8A8_SRGB texture for "FrontLoadingScreen"
set overlay R8G8B8A8_SRGB texture for "BackLoadingScreen"
set overlay R8G8B8A8_UNORM texture for "FrontLoadingScreenWorldImage"
set overlay R8G8B8A8_UNORM texture for "BackLoadingScreenWorldImage"
```
apparently the thumbnails are linear and our EColorSpace::Auto impl is not handling them correctly

I think it might be the case that the EColorSpace::Auto behavior is different in the case of Vulkan.

in the meantime, VRChat submits its R8G8B8A8_UNORM eye textures as EColorSpace::Auto and those look good, so I'm not sure what's happening there.

edit: ok i'm not going to touch it further because trying to fix the thumbnails breaks the eye texture gamma. don't want to go out of scope too much

myghi63:
hey, since it's working pretty well already, why not merge?

src/overlay.rs
```diff
@@ -286,21 +292,100 @@ impl OverlayMan {
...
+    /// `item` must be a non-null pointer to a valid XrBaseInStructure object
+    ///
+    /// SAFETY: For lifetime guarantees, store item in Box inside CompositorLayer.
+    unsafe fn next_chain_insert(&mut self, item: *mut xr::sys::BaseInStructure) {
+        let new_elem = item.as_mut().unwrap();
+        self.layer = Some(match self.layer.take().unwrap() {
+            OverlayLayerInner::Quad(quad) => {
+                let mut raw = quad.into_raw();
+                new_elem.next = raw.next as _;
+                raw.next = item.as_ref().unwrap() as *const _ as *const _;
```
Supreeeme, line 337: What's the point of going from a pointer to a reference and back to a pointer?

src/overlay.rs
```diff
@@ -286,21 +292,100 @@ impl OverlayMan {
...
+                OverlayLayerInner::Quad(xr::CompositionLayerQuad::<'a, G>::from_raw(raw))
```
Supreeeme, line 338: Could the turbofish be elided?
```suggestion
                OverlayLayerInner::Quad(xr::CompositionLayerQuad::from_raw(raw))
```

src/overlay.rs
galister: :facepalm:

galister:
rebased to main  not sure if auto-merge was stuck about that

PR #81, opened 2025-04-02
Handle gamma space textures on opengl backend

knah:
This fixes vivecraft on 1.20.4 being too bright. The behavior should to be in-line with descriptions of `EColorSpace`, although I'm not aware of any other games using OpenGL to verify possible breakage in the other direction.

### DIFF
--- a/src/graphics_backends/gl.rs
+++ b/src/graphics_backends/gl.rs
@@ -154,7 +154,7 @@ impl GraphicsBackend for GlData {
         &self,
         texture: Self::OpenVrTexture,
         bounds: vr::VRTextureBounds_t,
-        _color_space: vr::EColorSpace,
+        color_space: vr::EColorSpace,
     ) -> xr::SwapchainCreateInfo<Self::Api> {
         let mut fmt = 0;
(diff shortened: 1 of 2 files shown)


src/graphics_backends/gl.rs
```diff
@@ -163,6 +163,17 @@ impl GraphicsBackend for GlData {
...
+        let fmt = if is_srgb {
+            match fmt as gl::types::GLenum {
+                gl::RGBA8 => gl::SRGB8_ALPHA8 as gl::types::GLint,
+                gl::RGB8 => gl::SRGB8 as gl::types::GLint,
+                _ => fmt,
+            }
+        } else {
+            fmt
+        };
+
```
Supreeeme, line 176: Could just make this a single match.
```rust
let fmt = match color_space {
    vr::EColorSpace::Linear => fmt,
    vr::EColorSpace::Gamma | vr::EColorSpace::Auto => match fmt {
        // etc...
    }
};
```

knah: Done

PR #82, opened 2025-04-04
Add support for generic trackers using the XR_MNDX_xdev_space extension

### LINKED ISSUE #27: Introduce full body tracking using the XR_MNDX_xdev_space extension
Monado and WiVRn implement this OpenXR experimental extension to add support for arbitrary 6dof poses named and serialized per xdev frequently used for full body tracking on OpenComposite.

ricky26:
This is an implementation of generic trackers, using the `XR_MNDX_xdev_space` extension, similarly to OpenComposite. This is enough to get SteamVR-based full-body tracking in things like VRChat.

It's not quite there yet, but it does work: I've spent a few hours in VRChat with these changes (and the gesture changes in another PR).

Notes:
- We need an extension which isn't in-tree in OpenXRS. I've added out-of-tree bindings for it, but they're a little messy. There might be a better way to do this.
- I haven't implemented any events, or checking whether the device list has changed between frames. Theoretically this is supported by the API surface, but I don't think that this works in Monado yet (at least I don't think it does for SlimeVR).
- I'm not super happy with the changes needed in `end_session`, but apparently the cleanup order is important here: you have to destroy xdev-related resources before requesting session end.
- I only implemented pose fetching and string properties. It feels like there might be other facets of generic trackers I'm missing.

Theoretically this would fix #27. (I did see the 'dibs' in that issue, but I wanted this functionality and that was a couple of months ago.)

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -1,0 +1,336 @@
+// This file contains complete extension bindings, but only some of the bindings
+// are used.
+#![allow(dead_code)]
+
+use std::ffi::CStr;
+use std::mem;
+use std::os::raw::c_void;
+
+use openxr as xr;
+use openxr::sys as xrsys;
+
+fn convert_result(result: xrsys::Result) -> xr::Result<()> {
+    if result.into_raw() == 0 {
+        Ok(())
+    } else {
+        Err(result)
+    }
+}
+
+unsafe fn get_instance_proc_addr(
+    instance: &xr::Instance,
+    name: &CStr,
+) -> xr::Result<xrsys::pfn::VoidFunction> {
+    let mut f = None;
+    convert_result((instance.entry().fp().get_instance_proc_addr)(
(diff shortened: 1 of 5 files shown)


galister:
There's also an implementation here: https://github.com/RinLovesYou/xrizer/tree/experimental
(this was announced on the LVRA discord some weeks ago)

Which had https://github.com/Supreeeme/xrizer/pull/49 (that one was closed for it to be reopened after a rebase).

Might be worth to collab with Rin to get a best of both implementations.

RinLovesYou:
yeah after the device refactor is in, trackers were next on the list.

ricky26:
Ooh, I didn't know that branch existed (I didn't see the announcement on the LVRA Discord). I don't imagine there's anything worth taking from my implementation, it was fairly quick & dirty and written in a couple of hours because it didn't look like it was being worked on.

At least this will have left a GitHub paper trail connecting the issue to that branch.

Issue #84, opened 2025-04-18
[Bug]: No Input after initial Loading Screen in SkyrimVR


Erdragh:
### Operating System

Arch Linux

### OpenXR Runtime

Monado via Envision

### Describe the issue.

System:
- 7800x3d + 7900xt
- amdgpu + radv
- Steam Client Beta
- Oculus Rift S with Touch Controllers

Everything is set up via Envision. I can get input if I use the original OpenComposite, but that crashes when I have a mod like JContainers installed.

The weird thing is that I _do_ seem to get input during the initial loading screen (I can turn the 3d loading icon with the thumbsticks), but after that, in the "press any button" screen, it doesn't seem to be getting any.

Here's a log with `RUST_LOG=debug`:

[xrizer.txt](https://github.com/user-attachments/files/19813449/xrizer.txt)

### Steps to reproduce

1. Start an Envision Profile with latest xrizer
2. Start SkyrimVR
3. Try moving the loading icon
4. Observe not being able to pass the "press any button" screen.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Erdragh:
Looking through the log and the associated parts of the source code, could this be related to the game not receiving any supported buttons?
```
[2025-04-18T16:40:22.662 DEBUG xrizer::system ThreadId(2)] requesting uint64 property: SupportedButtons_Uint64 (1)
```
is all over the logs, but the respective implementation just _always_ returns `0`:
https://github.com/Supreeeme/xrizer/blob/bd7212552d6852752466e7f8a8b0f96fd4341598/src/system.rs#L517-L527

Supreeeme:
It's a good guess.

Erdragh:
After adding these, I can get past the "Press Any Button" Screen, but Skyrim shows me Vive Controllers instead of hands (which it showed me with OpenComposite). It also doesn't seem to be taking my thumbstick's input for anything other than the loading screens (so no menu navigation or movement) _and_ the buttons the Touch Controllers have, which the Vive Controllers don't, don't do anything either.

To me this seems like xrizer wrongly prioritizes legacy input here, whereas OpenComposite uses the actions provided by the game.

PR #85, opened 2025-04-20
Add legacy buttons mask

Erdragh:
SkyrimVR queries this when using legacy input doesn't allow passing the "Press Any Button" screen without this info. (It doesn't take input from buttons at all if it doesn't get this info)

This is part of #84.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1321,6 +1321,17 @@ impl<C: openxr_data::Compositor> Input<C> {
             _ => None,
         })
     }
+    
+    pub fn get_controller_uint_tracked_property(
+        &self,
+        hand: Hand,
+        property: vr::ETrackedDeviceProperty
+    ) -> Option<u64> {
+        self.get_profile_data(hand).and_then(|data| match property {
+            vr::ETrackedDeviceProperty::SupportedButtons_Uint64 => Some(data.legacy_buttons_mask),
+            _ => None
+        })
+    }
 
     pub fn post_session_restart(&self, data: &SessionData) {
         // This function is called while a write lock is called on the session, and as such should
--- a/src/input/profiles.rs
+++ b/src/input/profiles.rs
@@ -61,6 +61,9 @@ pub struct ProfileProperties {
     /// Can be found in SteamVR under resources/rendermodels (some are in driver subdirs)
     pub render_model_name: Property<&'static CStr>,
     pub main_axis: MainAxisType,
+    /// Corresponds to Prop_SupportedButtons_Uint64
+    /// Can (probably) be pulled from a SteamVR System Report
+    pub legacy_buttons_mask: u64
 }
 
 pub(super) struct PathTranslation {
--- a/src/input/profiles/knuckles.rs
+++ b/src/input/profiles/knuckles.rs
@@ -22,6 +22,7 @@ impl InteractionProfile for Knuckles {
                 right: c"valve_controller_knu_1_0_right",
             },
             main_axis: MainAxisType::Thumbstick,
+            legacy_buttons_mask: (1 << 0) | (1 << 1) | (1 << 2) | (1 << 7) | (1 << 32) | (1 << 33) | (1 << 34)
         }
     }
     fn translate_map(&self) -> &'static [PathTranslation] {
--- a/src/input/profiles/oculus_touch.rs
+++ b/src/input/profiles/oculus_touch.rs
@@ -18,6 +18,7 @@ impl InteractionProfile for Touch {
                 right: c"oculus_quest_controller_right",
             },
             main_axis: MainAxisType::Thumbstick,
+            legacy_buttons_mask: (1 << 0) | (1 << 1) | (1 << 2) | (1 << 7) | (1 << 32) | (1 << 33) | (1 << 34),
         }
     }
     fn profile_path(&self) -> &'static str {
--- a/src/input/profiles/simple_controller.rs
+++ b/src/input/profiles/simple_controller.rs
@@ -16,6 +16,7 @@ impl InteractionProfile for SimpleController {
             openvr_controller_type: c"<unknown>",
             render_model_name: Property::BothHands(c"generic_controller"),
             main_axis: MainAxisType::Thumbstick,
+            legacy_buttons_mask: (1 << 0) | (1 << 1) | (1 << 2) | (1 << 32) | (1 << 33) // TODO: This is just the one for the vive_controller. I'm not certain whether that's correct here
         }
     }
     fn profile_path(&self) -> &'static str {
--- a/src/input/profiles/vive_controller.rs
+++ b/src/input/profiles/vive_controller.rs
@@ -15,6 +15,7 @@ impl InteractionProfile for ViveWands {
             openvr_controller_type: c"vive_controller",
             render_model_name: Property::BothHands(c"vr_controller_vive_1_5"),
             main_axis: MainAxisType::Trackpad,
+            legacy_buttons_mask: (1 << 0) | (1 << 1) | (1 << 2) | (1 << 32) | (1 << 33)
         }
     }
     fn profile_path(&self) -> &'static str {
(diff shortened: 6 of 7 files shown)


src/input/profiles/knuckles.rs
```diff
@@ -22,6 +22,7 @@ impl InteractionProfile for Knuckles {
                 right: c"valve_controller_knu_1_0_right",
             },
             main_axis: MainAxisType::Thumbstick,
+            legacy_buttons_mask: (1 << 0) | (1 << 1) | (1 << 2) | (1 << 7) | (1 << 32) | (1 << 33) | (1 << 34)
```
Supreeeme, line 25: These should use the button ids (`vr::EVRButtonId`), and perhaps the `button_mask_from_id` function in `legacy.rs`.

Erdragh: I've now extracted the `ProfileProperty`s into `static` fields and made the mentioned function `pub const` to be able to use them.
It now uses the verbose names for the buttons, so it's much more readable

Red-M:
I've tested this and made an extension to the patch (which is in a PR in the fork), it does allow games to make better choices on what kind of input device you've got, in my case, Resonite actually detects my knuckles properly now.

src/input.rs
```diff
@@ -1283,14 +1283,16 @@ impl<C: openxr_data::Compositor> Input<C> {
                 }
                 // I Expect You To Die 3 identifies controllers with this property -
                 // why it couldn't just use ControllerType instead is beyond me...
-                vr::ETrackedDeviceProperty::ModelNumber_String => Some(data.model),
+                vr::ETrackedDeviceProperty::ModelNumber_String => Some(data.model), // TODO: This seems weird
```
Supreeeme, line 1286:
```suggestion
                vr::ETrackedDeviceProperty::ModelNumber_String => Some(data.model),
```
It is what it is

src/input/profiles.rs
```diff
@@ -61,6 +61,17 @@ pub struct ProfileProperties {
...
+    /// Corresponds to Prop_RegisteredDeviceType_String
+    pub registered_device_type: Property<&'static CStr>,
+    /// Corresponds to Prop_SerialNumber_String
+    pub serial_number: Property<&'static CStr>,
+    /// Corresponds to Prop_TrackingSystemName_String
+    pub tracking_system_name: &'static CStr,
+    /// Corresponds to Prop_ManufacturerName_String
+    pub manufacturer_name: &'static CStr,
+    /// Corresponds to Prop_SupportedButtons_Uint64
+    /// Can (probably) be pulled from a SteamVR System Report
```
Supreeeme, line 73:
```suggestion
    /// Can be pulled from a SteamVR System Report
```

src/input/profiles/knuckles.rs
```diff
@@ -1,28 +1,43 @@
...
-        &ProfileProperties {
-            model: c"Knuckles",
-            openvr_controller_type: c"knuckles",
-            render_model_name: Property::PerHand {
-                left: c"valve_controller_knu_1_0_left",
-                right: c"valve_controller_knu_1_0_right",
-            },
-            main_axis: MainAxisType::Thumbstick,
-        }
+        &DEVICE_PROPERTIES
```
Supreeeme, line 40: This is already a static struct, there's no need to pull it out of the function.
Also, statics can live inside of functions anyway.

src/input/profiles/oculus_touch.rs
```diff
@@ -2,23 +2,40 @@ use super::{
...
+        right: c"oculus/WMHD315M3010GV_Controller_Right"
+    },
+    serial_number: Property::PerHand {
+        left: c"WMHD315M3010GV_Controller_Left",
+        right: c"WMHD315M3010GV_Controller_Right"
+    },
+    tracking_system_name: c"oculus",
+    manufacturer_name: c"Oculus",
+    main_axis: MainAxisType::Thumbstick,
+    //(1 << 0) | (1 << 1) | (1 << 2) | (1 << 7) | (1 << 32) | (1 << 33) | (1 << 34),
```
Supreeeme, line 32:
```suggestion
```

src/input/profiles/simple_controller.rs
```diff
@@ -1,22 +1,39 @@
...
+        left: c"htc/vive_controllerLHR-00000001",
+        right: c"htc/vive_controllerLHR-00000002"
+    },
+    serial_number: Property::PerHand {
+        left: c"LHR-00000001",
+        right: c"LHR-00000002"
+    },
+    tracking_system_name: c"lighthouse",
+    manufacturer_name: c"HTC",
+    //(1 << 0) | (1 << 1) | (1 << 2) | (1 << 32) | (1 << 33)
```
Supreeeme, line 30:
```suggestion
```

src/input/profiles/vive_controller.rs
```diff
@@ -2,20 +2,37 @@ use super::{
...
+        left: c"htc/vive_controllerLHR-00000001",
+        right: c"htc/vive_controllerLHR-00000002"
+    },
+    serial_number: Property::PerHand {
+        left: c"LHR-00000001",
+        right: c"LHR-00000002"
+    },
+    tracking_system_name: c"lighthouse",
+    manufacturer_name: c"HTC",
+    //(1 << 0) | (1 << 1) | (1 << 2) | (1 << 32) | (1 << 33)
```
Supreeeme, line 29:
```suggestion
```

Supreeeme (commented):
How did you get the properties for all of these controllers?

Erdragh:
> How did you get the properties for all of these controllers?

They're taken from Valve's docs on emulatable devices (https://github.com/ValveSoftware/openvr/blob/master/docs/Driver_API_Documentation.md#emulateable-devices).

On the Discord, it was stated that these aren't necessarily reliable and a link to OC's values was shared, so I'll compare and contrast those. I'll also look at merging the above mentioned PR into my branch too.

Erdragh:
> On the Discord, it was stated that these aren't necessarily reliable and a link to OC's values was shared, so I'll compare and contrast those.

Having looked at OC's strings here, they are mostly the same as the ones from Valve's docs, with my PR having extra properties that aren't in OC (e.g. serial number). The only difference I could find was the touch controllers, where OC seems to use the quest 2's touch controller properties, whereas the valve docs document the quest 3's touch controllers. OC also defaults the RegisteredDeviceType to the Oculus controllers, which this PR doesn't.

src/input/profiles/knuckles.rs
Erdragh: Ok, I'll put the `static`s inside the function. This can't work without the keyword because using the `button_mask_from_id` function will result in "Cannot return reference to temporary value".

src/input/profiles.rs
```diff
@@ -53,14 +53,25 @@ pub enum MainAxisType {
...
     /// Corresponds to RenderModelName_String
     /// Can be found in SteamVR under resources/rendermodels (some are in driver subdirs)
     pub render_model_name: Property<&'static CStr>,
     pub main_axis: MainAxisType,
+    /// Corresponds to Prop_RegisteredDeviceType_String
+    pub registered_device_type: Property<&'static CStr>,
+    /// Corresponds to Prop_SerialNumber_String
+    pub serial_number: Property<&'static CStr>,
+    /// Corresponds to Prop_TrackingSystemName_String
+    pub tracking_system_name: &'static CStr,
```
Supreeeme, line 69: I wonder, is registered_device_type and tracking_system_name actually needed for functionality? I'd prefer to avoid providing properties that games won't actually use for something meaningful.

Erdragh: `tracking_system_name` is what Skyrim VR seems to actually need, but the logs show it also asks for the registered device type.

Red-M: I think in this particular instance of exposing device information, its not about if a game uses it but if a game COULD read them previously, then it should because any game could read these and make different choices based on the fact it had access to them in OpenVR.

While this might not match OpenXR's or OpenVR's specifications, the developers of the game did and thats what users are going to see the results of.

Red-M:
What can we do from here to get this merged into `main`?

PR #94, opened 2025-04-29
input: fix input with only one controller connected

openglfreak:


RinLovesYou:
this is actually fixed in #68

Supreeeme:
Given that I see #68 taking a while I don't mind merging this in the meantime.
This could use a test.

openglfreak:
It looks like this is fixed on master, closing.

PR #109, opened 2025-05-30
[WIP] App bootstrapper

oo8dev:
Essentially, the plan is to launch user-specified applications once an OpenVR bootstrapper client connects to xrizer.

cc @SpookySkeletons

### DIFF
--- a/src/clientcore.rs
+++ b/src/clientcore.rs
@@ -116,6 +116,33 @@ impl ClientCore {
         .unwrap();
         Some(get_interface(&item))
     }
+
+    // Example usage:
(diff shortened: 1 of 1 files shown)


Supreeeme:
What's the point of this exactly? What clients are even an "OpenVR bootstrapper"?

SpookySkeletons:
Originally the steamvr button in the top right of the steam client would call the steam bootstrapper OpenVR client which would open steamvr and all the rest but it appears this behavior has changed and it now opens steamvr directly, the goal here was a bit better integration by kicking off the likes of wlx but I guess that's not happening now after whatever change occurred.

PR #114, opened 2025-06-11
Screenshots interface to return IncompatibleVersion instead of panic

galister:
There are titles (Basis) that attempt to use this interface. Since there's even an appropriate error code, let's fail in a proper manner that the app can then handle.

### DIFF
--- a/src/screenshots.rs
+++ b/src/screenshots.rs
@@ -15,7 +15,7 @@ impl vr::IVRScreenshots001_Interface for Screenshots {
         _: *const std::os::raw::c_char,
         _: *const std::os::raw::c_char,
     ) -> vr::EVRScreenshotError {
-        todo!()
+        vr::EVRScreenshotError::IncompatibleVersion
     }
     fn TakeStereoScreenshot(
         &self,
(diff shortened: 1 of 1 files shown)


src/screenshots.rs
```diff
@@ -15,39 +15,41 @@ impl vr::IVRScreenshots001_Interface for Screenshots {
...
         &self,
         _: vr::ScreenshotHandle_t,
         _: vr::EVRScreenshotPropertyFilenames,
         _: *mut std::os::raw::c_char,
         _: u32,
-        _: *mut vr::EVRScreenshotError,
+        err: *mut vr::EVRScreenshotError,
     ) -> u32 {
-        todo!()
+        unsafe { *err = vr::EVRScreenshotError::IncompatibleVersion };
```
Supreeeme: Should have a null check

src/screenshots.rs
```diff
@@ -15,39 +15,41 @@ impl vr::IVRScreenshots001_Interface for Screenshots {
...
+        0
     }
     fn GetScreenshotPropertyType(
         &self,
         _: vr::ScreenshotHandle_t,
-        _: *mut vr::EVRScreenshotError,
+        err: *mut vr::EVRScreenshotError,
     ) -> vr::EVRScreenshotType {
-        todo!()
+        unsafe { *err = vr::EVRScreenshotError::IncompatibleVersion };
```
Supreeeme: Same here

galister:
added null checks.

Issue #122, opened 2025-06-30
[Bug]: IVROverlay_IVROverlay_020_SetOverlayInputMethod not yet implemented [The LAB]


Mr-Zero88:
### Operating System

Arch Linux

### OpenXR Runtime

WiVRn

### Describe the issue.

The Lab crashes after the initial loding screen.

```
[2025-06-30T02:55:46.018 INFO  xrizer ThreadId(1)] Initializing XRizer
[2025-06-30T02:55:46.018 INFO  xrizer::clientcore ThreadId(1)] Creating ClientCore version "IVRClientCore_003"
[2025-06-30T02:55:46.205 INFO  xrizer::openxr_data ThreadId(1)] New session created!
[2025-06-30T02:55:46.205 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: READY
[2025-06-30T02:55:46.205 INFO  xrizer::openxr_data ThreadId(1)] Began OpenXR session.
[2025-06-30T02:55:46.205 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: SYNCHRONIZED
[2025-06-30T02:55:46.205 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: STOPPING
[2025-06-30T02:55:46.205 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: IDLE
[2025-06-30T02:55:46.205 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: EXITING
[2025-06-30T02:55:47.612 INFO  xrizer::clientcore ThreadId(1)] Creating ClientCore version "IVRClientCore_003"
[2025-06-30T02:55:47.793 INFO  xrizer::openxr_data ThreadId(1)] New session created!
[2025-06-30T02:55:47.793 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: READY
[2025-06-30T02:55:47.793 INFO  xrizer::openxr_data ThreadId(1)] Began OpenXR session.
[2025-06-30T02:55:47.793 WARN  xrizer::clientcore ThreadId(1)] app requested unknown interface "IVRChaperoneSetup_005"
[...]

PR #126, opened 2025-06-30
Fix IVROverlay_IVROverlay_020_SetOverlayInputMethod not yet implemented [The LAB]

Mr-Zero88:
Fixes: https://github.com/Supreeeme/xrizer/issues/122
Adds IVROverlay_IVROverlay_020_SetOverlayInputMethod with an unimplemented warning with the two types.
Adds the Tip type used in the action profile of The Lab for Oculus touch controllers

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -275,7 +275,13 @@ struct BoundPose {
 enum BoundPoseType {
     /// Equivalent to what is returned by WaitGetPoses, this appears to be the same or close to
     /// OpenXR's grip pose in the same position as the aim pose.
+    /// "All tracked devices also get two pose components registered regardless of what render model they use: /pose/raw and /pose/tip"
+    /// "By default, both are set to the unaltered pose of the device."
+    /// ~https://github.com/ValveSoftware/openvr/wiki/Input-Profiles#pose-components
     Raw,
+    /// "If you provide /pose/tip in your rendermodel you should set it to the position and rotation that are appropriate for pointing (i.e. with a laser pointer) with your controller."
+    /// ~https://github.com/ValveSoftware/openvr/wiki/Input-Profiles#pose-components
+    Tip,
     /// Not sure why games still use this, but having it be equivalent to raw seems to work fine.
     Gdc2015,
 }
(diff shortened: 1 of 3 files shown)


src/input/action_manifest.rs
```diff
@@ -533,6 +533,7 @@ fn parse_pose_binding<'de, D: serde::Deserializer<'de>>(
 
     let pose = match pose {
         "raw" => BoundPoseType::Raw,
+        "tip" => BoundPoseType::Tip,
         "gdc2015" => BoundPoseType::Gdc2015,
         other => return Err(D::Error::unknown_variant(other, &["raw", "gdc2015"])),
```
Supreeeme:
```suggestion
        other => return Err(D::Error::unknown_variant(other, &["raw", "tip", "gdc2015"])),
```

src/input.rs
```diff
@@ -770,7 +771,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
                 });
 
                 match ty {
-                    BoundPoseType::Raw | BoundPoseType::Gdc2015 => (origin, hand),
+                    BoundPoseType::Raw | BoundPoseType::Tip | BoundPoseType::Gdc2015 => (origin, hand),
```
Supreeeme: I am not so sure that Tip and Raw should be the same pose, but I suppose we can cross that bridge when we get to it.

Supreeeme:
Be sure to run `cargo fmt` and `cargo clippy` on the repo.

src/input.rs
Mr-Zero88: "All tracked devices also get two pose components registered regardless of what render model they use: /pose/raw and /pose/tip. By default, both are set to the unaltered pose of the device. If you provide /pose/tip in your rendermodel you should set it to the position and rotation that are appropriate for pointing (i.e. with a laser pointer) with your controller." ~https://github.com/ValveSoftware/openvr/wiki/Action-manifest#default_bindings

So the default for "/tip" is the "/raw" position. If the current render model defines a "/tip" component, than we should apply the component_local transform.

But for now this also works with the only different being that the start pose and rotation may be off by a small amount.

Like for quest 2:

```JSON
{
  "origin": [0.016694, -0.02522, 0.024687],
  "rotate_xyz" : [-37.4,0.0,0.0]
}
```

The position is around the edge of the controller ring and rotated 37° up.

Mr-Zero88:
`cargo clippy` throws a bunch of `variables can be used directly in the 'format!' string` and `using 'contains()' instead of 'iter().any()' is more efficient` errors in places I have not touched.

src/input.rs
Supreeeme: 😮 nice find. Would you mind adding a comment on the `Tip` variant on the BoundPoseType enum, and add that link?

Supreeeme:
Looks good, could you squash the last couple commits into the one adding Tip? So there's only two commits.

PR #130, opened 2025-07-06
Add support for Windows Mixed Reality controllers

CheesecakeCG:
I've only added support for the original WMR controllers so far, and I haven't much tested outside of VRChat and ChilloutVR.

~~Also, the grip poses in Monado might be flipped... and the ones I'm using here are rotated wrong.~~
The offsets are now fixed in Beyley's `rift-driver-tracking` branch and Thaytan's `dev-constellation-controller-tracking` branch of Monado.

- [x] OG WMR Controller Support (mostly done)
- [x] Samsung Odyssey Controller Support
- [X] HP Reverb G2 Controller Support

I'll need other people to test the last two controllers.

### DIFF
--- a/src/input/profiles.rs
+++ b/src/input/profiles.rs
@@ -1,10 +1,18 @@
 pub mod knuckles;
+
 pub mod oculus_touch;
+
 pub mod simple_controller;
 pub mod vive_controller;
+
+pub mod wmr;
+
+
 pub mod vive_focus3;
+
 #[cfg(feature = "monado")]
 pub mod vive_tracker;
+
 use super::{
     action_manifest::ControllerType, legacy::LegacyBindings, skeletal::SkeletalInputBindings,
 };
@@ -17,6 +25,11 @@ use simple_controller::SimpleController;
 use std::ffi::CStr;
 use vive_controller::ViveWands;
 use vive_focus3::ViveFocus3;
+use wmr::{
+    hp_motion_controller::ReverbG2Controller, ms_motion_controller::HolographicController,
+    samsung_odyssey_controller::SamsungOdysseyController,
+};
+
 
 #[allow(private_interfaces)]
 pub trait InteractionProfile: Sync + Send {
@@ -113,6 +126,15 @@ impl Profiles {
         // Add supported interaction profiles here.
         static P: Profiles = Profiles {
             list: &[
+                (ControllerType::HPMotionController, &ReverbG2Controller),
+                (
+                    ControllerType::SamsungOdysseyController,
+                    &SamsungOdysseyController,
+                ),
+                (
+                    ControllerType::HolographicController,
+                    &HolographicController,
+                ),
                 (ControllerType::ViveController, &ViveWands),
                 (ControllerType::Knuckles, &Knuckles),
                 (ControllerType::OculusTouch, &Touch),
--- a/src/input/profiles.rs
+++ b/src/input/profiles.rs
@@ -1,18 +1,11 @@
 pub mod knuckles;
-
 pub mod oculus_touch;
-
 pub mod simple_controller;
 pub mod vive_controller;
-
-pub mod wmr;
-
-
 pub mod vive_focus3;
-
 #[cfg(feature = "monado")]
 pub mod vive_tracker;
-
+pub mod wmr;
 use super::{
     action_manifest::ControllerType, legacy::LegacyBindings, skeletal::SkeletalInputBindings,
 };
@@ -29,7 +22,6 @@ use wmr::{
     hp_motion_controller::ReverbG2Controller, ms_motion_controller::HolographicController,
     samsung_odyssey_controller::SamsungOdysseyController,
 };
-
 
 #[allow(private_interfaces)]
 pub trait InteractionProfile: Sync + Send {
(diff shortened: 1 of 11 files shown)


Supreeeme:
Feel free to split odyssey and reverb g2 support into separate PRs, if you'd like.

src/input/profiles.rs
```diff
@@ -1,14 +1,21 @@
...
 pub mod simple_controller;
 pub mod vive_controller;
 
 use super::{
     action_manifest::ControllerType, legacy::LegacyBindings, skeletal::SkeletalInputBindings,
 };
-use crate::openxr_data::Hand;
+use crate::{
+    input::profiles::samsung_odyssey_controller::SamsungOdysseyController, openxr_data::Hand,
+};
```
Supreeeme:
```suggestion
use crate::openxr_data::Hand;
use samsung_odyssey_controller::SamsungOdysseyController;
```

src/input/profiles.rs
```diff
@@ -109,6 +116,15 @@ impl Profiles {
         // Add supported interaction profiles here.
         static P: Profiles = Profiles {
             list: &[
+                (ControllerType::HPMotionController, &ReverbG2Controller),
+                (
+                    ControllerType::SamsungOdysseyController,
+                    &SamsungOdysseyController,
+                ),
```
Supreeeme, line 125: Since these controller types are locked behind extensions, we will need to handle the extensions not being available in some way.

Supreeeme (commented):
These could use some simple `verify_bindings` tests. You can look at the ones in the other profiles for inspiration.

toyowheelin:
I am not sure if this is a good place for this or not but I built Cheesecakes wmr-controller-support branch and tested it with Xplane11 and ended up with a crash. I was in communication with Supreme in Discord and they were thinking it was likely that Cheesecakes branch just needs to be rebased. I would love to test and am hopeful that Cheesecake will see this and be able to do that. For reference this is the log from the crash https://bin.garudalinux.org/?d2bb0dfa11ef9f74#7uRhUQbjF928BVyHuLm8km9aXs7Cp8WRc6LBzE2YspYC

toyowheelin:
Unfortunately the rebase didn't help and resulted in the same error. https://bin.garudalinux.org/?11a46fbeba5363eb#BAPbEdpGxLu6gvtc8L7AZ5KZQaoYLyPU7XiYnNJQuG6 I will try to compile the default xrizer and make sure it works, as I have only ever used OpenComposite with Xplane11.
EDIT: Good/Bad news, default xrizer works great. So something is broken on the wmr-controller-support branch I guess. Log from default xrizer https://bin.garudalinux.org/?6166151fa5d4a6bc#BfL5Wfp4FAaJo8FUqNQDSAmJCAsG68yZQGzaBUvdynLt

CheesecakeCG:
@toyowheelin Which kind of controllers are you using? Reverb G2?

toyowheelin:
I have a Samsung Odyssey + setup.

toyowheelin:
Let me know if you need any more info to look into this, I would love to help make this work.

CheesecakeCG:
> Let me know if you need any more info to look into this, I would love to help make this work.

I still need the render_model_name for the Samsung controllers, unless SteamVR doesn't actually show different models for those controllers. I'm not entirely sure where in the SteamVR driver that's defined though, I'm able to see it in the other drivers but not for WMR.

Supreeeme:
It'd be nice to have these in, if you want to move Samsung support to a separate PR.

src/input/profiles.rs
Supreeeme: #298 adds a mechanism for this, so you can use that once that gets merged.

src/input/profiles/wmr/hp_motion_controller.rs
```diff
@@ -0,0 +1,244 @@
...
+    fn properties(&self) -> &'static ProfileProperties {
+        static DEVICE_PROPERTIES: ProfileProperties = ProfileProperties {
+            model: Property::PerHand {
+                left:c"WindowsMR: 0x045E/0x066A/0/1",
+                right: c"WindowsMR: 0x045E/0x066A/0/2",
+            },
+            openvr_controller_type: c"hpmotioncontroller",
+            render_model_name: Property::PerHand {
+                left: c"C:\\Users\\steamuser\\AppData\\Local\\Microsoft/Windows/OpenVR\\controller_1642_1118_1\\controller.obj",
+                right: c"C:\\Users\\steamuser\\AppData\\Local\\Microsoft/Windows/OpenVR\\controller_1642_1118_2\\controller.obj"
```
Supreeeme, line 29: I assume this is what's returned in Windows? Kind of cursed...

CheesecakeCG: It's pretty close to what's returned on Windows, except the user name usually matches the actual user's username. I figured most of the time, people will be running games with XRizer through Proton and that's the username used there.

src/input/profiles.rs
Supreeeme: Now merged!

CheesecakeCG:
~~I think the only issue left~~ (unless another problem comes up when testing with Reverb G2 controllers) ~~is that on the OpenVR side of things there's no distinction between the original WMR controllers and the Samsung reshelled version besides the render model and offsets.~~

~~Using `#[serde(rename = "holographic_controller")]` on SamsungOdysseyController `action_manifest::ControllerType` just makes serde unhappy...~~

I also squashed all the old commits to make rebasing easier.

Edit: I think I fixed the issue with Samsung controllers. Now it just needs users to test it.

Supreeeme:
Sorry, I haven't mentioned it, but I overhauled how input profiles are implemented. Hopefully not too much work to rebase.

PR #176, opened 2025-08-26
Add support for Half-Life 2 VR

Mr-Zero88:
- Add stubs for dominant hand handling
- Implement GetLastPoseForTrackedDeviceIndex with pose handling for tracked devices.
- Add new BoundPoseType variants and update pose parsing logic to comply with SteamVR default
- Refactor pose retrieval to use a loop for all devices with Input::get_device_pose

PR #182, opened 2025-09-03
Just enough to get HL2VR to start

ada-tv:
This resolves startup panics and looks to get poses working.

I can't quite test this with my real hardware, I'm still trying to figure out how to get the 32-bit stuff working properly. Monado's simulated HMD view looks correct though.

image

Should also fix a panic in Half-Life: Alyx in `SetDominantHand` when the dominant hand setting is changed. The game has its own internal dominant hand setting, so I think it just calls that to hint SteamVR to load the left/right-handed bindings.

`GetLastPoseForTrackedDeviceIndex` was scavenged from #176 and modified to not need extra helper functions.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -795,25 +795,30 @@ impl vr::IVRCompositor028_Interface for Compositor {
             return vr::EVRCompositorError::RequestFailed;
         }
 
-        // HMD, left hand, right hand
-        let mut poses = [Default::default(); 3];
-        self.input
-            .force(|_| Input::new(self.openxr.clone()))
-            .get_poses(&mut poses, None);
+        let input = self.input.force(|_| Input::new(self.openxr.clone()));
 
-        if let Some(pose) = poses.get(device_index as usize) {
-            if !output_pose.is_null() {
-                unsafe { output_pose.write(*pose) };
+        let pose = match device_index {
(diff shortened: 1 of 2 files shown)


src/compositor.rs
```diff
@@ -787,11 +787,33 @@ impl vr::IVRCompositor028_Interface for Compositor {
...
-        todo!()
+        if output_pose.is_null() && output_game_pose.is_null() {
+            return vr::EVRCompositorError::RequestFailed;
+        }
+
+        // HMD, left hand, right hand
+        let mut poses = [Default::default(); 3];
+        self.input
+            .force(|_| Input::new(self.openxr.clone()))
+            .get_poses(&mut poses, None);
```
Supreeeme: This is grabbing 3 poses, but we only need to grab one.

Mr-Zero88:
I will try to pick all the commits and make a proper PR once I fixed the anoying crash after some time. Does the crash not happen with this patch?

ada-tv:
HL2VR doesn't crash on startup with this PR, if there's a crash that happens later I haven't seen it.

Mr-Zero88:
Yes with my patches you can play the intro for like 5 - 10 minutes when the game just exits without errors.

PR #192, opened 2025-09-09
No man sky fixes

Mr-Zero88:
Fix some crashes in no man sky

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -1843,6 +1843,20 @@ mod tests {
     }
 
     #[test]
+    fn explicit_timing_multiple_submit_explicit_timing_data() {
+        let f = Fixture::new();
+        f.ensure_real_session(false);
+        f.comp.SetExplicitTimingMode(
+            vr::EVRCompositorTimingMode::Explicit_ApplicationPerformsPostPresentHandoff,
+        );
+
+        assert_eq!(f.comp.SubmitExplicitTimingData(), None);
+        f.check_frame_state(fakexr::FrameState::Begun);
+        assert_eq!(f.comp.SubmitExplicitTimingData(), None);
+        f.check_frame_state(fakexr::FrameState::Begun);
+    }
+
+    #[test]
     fn explicit_timing_session_restart_after_waitgetposes() {
         let f = Fixture::new();
         f.comp.SetExplicitTimingMode(


src/compositor.rs
```diff
@@ -47,24 +47,31 @@ enum FrameState {
 }
 
 impl FrameState {
-    fn advance_to(&mut self, new: Self) {
+    fn advance_to(&mut self, new: Self) -> Option<Self> {
```
Supreeeme:
```suggestion
    fn advance_to(&mut self, new: Self) -> bool {
```

src/compositor.rs
```diff
@@ -47,24 +47,31 @@ enum FrameState {
...
         let old = *self;
-        let mut allowed = |a| {
-            assert!(new == a, "Tried to advance from {:?} to {new:?}", *self);
+        let mut allowed = |a| -> Option<FrameState> {
+            if new != a {
+                info!(
+                    "Tried to advance from {:?} to {new:?} which is not allowed",
+                    *self
+                );
+                return None;
```
Supreeeme:
```suggestion
                warn!(
                    "Tried to advance from {:?} to {new:?}",
                    *self
                );
                return false;
```

Mr-Zero88:
Now restricted FrameState to not transition from Begun to Begun again, as that would cause a ERROR_CALL_ORDER_INVALID exception in the graphics pipeline.

src/compositor.rs
```diff
@@ -47,24 +47,26 @@ enum FrameState {
...
-            }
-            Self::Begun => *self = new,
-            Self::Submitted => {
-                allowed(Self::Waited);
-            }
-        }
+        let result: bool = match old {
+            Self::Waited => allowed(vec![Self::Begun]),
+            Self::Begun => allowed(vec![Self::Waited, Self::Submitted]),
+            Self::Submitted => allowed(vec![Self::Waited]),
```
Supreeeme: A Vec seems like overkill here imo. I would just change the `Self::Begun` branch to check that `new != Self::Begun`.

Supreeeme (commented):
The `advance_to` change could use a test. This would be triggered by calling `SubmitExplicitTimingData` twice in succession.

Mr-Zero88:
> The `advance_to` change could use a test. This would be triggered by calling `SubmitExplicitTimingData` twice in succession.

Added the test case.

PR #194, opened 2025-09-09
Basis Temp fix: add workaround for Unity 6.2 path bug causing buffer overflow

Mr-Zero88:
Unity 6.2 has introduced a bug where they pass a path with doubled instalation paths and a buffer overflow caused by a missing null terminator.
This handles this behavior.


Supreeeme:
Does this not break on SteamVR?

Mr-Zero88:
Just tried it and yes, it does also crash in SteamVR. So this would make xrizer the only way to play Unity 6.2 games.

Supreeeme:
I am not sure about adding a workaround for this, it sounds like Unity/Basis should just be fixed, especially if it's breaking SteamVR.

galister:
Unity's SteamVR plugin is open source and available here: https://github.com/ValveSoftware/steamvr_unity_plugin

I would wager that the Basis devs can just patch the relevant code in the plugin itself, since any and all OpenVR calls are made from this plugin.

PR #202, opened 2025-09-20
Link to system OpenXR instead of static compiling

SpookySkeletons:
To my awareness, Proton should now deal with this correctly, as should we.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ crate-type = ["cdylib"]
 
 [features]
 default = ["monado"]
+static-openxr = ["openxr/static"]
 tracing = ["dep:tracy-client", "openvr/tracing"]
 monado = ["dep:openxr_mndx_xdev_space"]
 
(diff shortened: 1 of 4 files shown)


SpookySkeletons:
Modified the CI to also pull in the OpenXR 1.0.20 headers for the build.

Supreeeme:
Thinking about this, I think I'd like if this was behind a feature (that can be on by default).

SpookySkeletons:
Feature gated and defaulted on

SpookySkeletons:
CI refusing external changes to the build.

Cargo.toml
```diff
@@ -13,6 +13,8 @@ readme = "README.md"
 crate-type = ["cdylib"]
 
 [features]
+default = ["openxr-dynamic"]
+openxr-dynamic = ["openxr/linked", "openvr/openxr-dynamic"]
```
Supreeeme: With the flags like this it's impossible to get static openxr anymore. It seems that the `static` flag actually implies `linked`, so maybe it should instead be a `openxr-static` feature and set `features = ["linked"]` on the openxr dependency.

SpookySkeletons: I believe that should do it now. Regular builds are linked and the CI remains static. Tests pass locally. When the AUR or distribution packaging pulls this in it will default to linking.

PR #204, opened 2025-09-23
Add openvr 2.12.14 [No man's sky 6.05]

Mr-Zero88:
Add IVRSystem023, IVROverlay028 and IVRCompositor029 for openvr 2.12.14

Update IVROverlay interface to implement CreateSubviewOverlay and SetSubviewPosition methods

### DIFF
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -792,7 +792,8 @@ impl vr::IVROverlay028_Interface for OverlayMan {
         _: *const c_char,
         _: u64,
     ) -> vr::EVROverlayError {
-        todo!()
+        crate::warn_unimplemented!("ShowKeyboardForOverlay");
+        vr::EVROverlayError::RequestFailed
     }
     fn ShowKeyboard(
         &self,


myghi63:
I've just tested and I was able to play No Man's Sky fine with this branch no-man-sky-fixes. The only Issue I've found is that all buttons are shown as " :warning: " in exception of thumbsticks:
Screenshot_org meumeu wivrn github testing_2025 09 23-13 02 27 702_267

Mr-Zero88:
Ok yes that is next on my watch of problems as that is also needed for HalfLife2VR mod. But then all the needed fixes for the new version are done.
Thanks for testing this again : )

src/system.rs
```diff
@@ -449,6 +452,34 @@ impl vr::IVRSystem022_Interface for System {
...
+        if overlay_handle.is_null() {
+            return false;
+        };
+        let handle = unsafe { overlay_handle.read() };
+        if vr::IVROverlay027_Interface::PollNextOverlayEvent(overlay.as_ref(), handle, event, size)
+        {
+            let event_data = unsafe { event.as_ref().unwrap().data };
+            unsafe { overlay_handle.write(event_data.overlay.overlayHandle) };
+            return true;
+        }
```
Supreeeme, line 479: I am pretty sure `overlay_handle` is not going to be a valid overlay handle, but something that xrizer is supposed to fill. I don't think you can have a meaningful implementation here until there's a real implementation of `PollNextOverlayEvent`

Mr-Zero88: I analysed places where that function is used, and they either give an overlay handle to specify a restriction or supply defalut 0 aka. InvalidHandle, which means give me any new event from any overlay. Then, the handle gets set to the overlay from which the returned event is from.

Supreeeme: What do you mean, what did you analyze?

Supreeeme: If I look at gamescope's use of this function: https://github.com/ValveSoftware/gamescope/blob/0d3a19791de447a673e315a5aed3f59ad64ebb71/src/Backends/OpenVRBackend.cpp#L1018

It is not giving a valid overlay handle. It's giving a pointer to an overlay handle that should be filled by the runtime if the event that's returned is actually an overlay event.

Mr-Zero88: That use case is already implemented in the above code. If an empty handle(0) is supply it will not filter the overlay and query all for events.

Supreeeme: My point is the application is not actually passing an overlay handle, it's passing a pointer to one that xrizer should fill.

Supreeeme:
Merged with an adjusted change to `PollNextEventWithPoseAndOverlays`.

PR #210, opened 2025-10-04
Fill OpenXR engine name/version

xytovl:
Use XRizer as the engine name, compute version from cargo metadata

Engine name and versions can be used by the runtime to enable some quirks. We don't have any for XRizer, but it might be useful at some point.

Making draft before we resolve a few points:
* Is the actual engine available from OpenVR?
* The cargo version is currently 0.3.0, is there a way to get the number of commits since the tag?

### DIFF
--- a/src/openxr_data.rs
+++ b/src/openxr_data.rs
@@ -8,6 +8,7 @@ use glam::f32::{Quat, Vec3};
 use log::{info, warn};
 use openvr as vr;
 use openxr as xr;
+use pkg_version::*;
 use std::mem::ManuallyDrop;
 use std::sync::{
     atomic::{AtomicBool, AtomicI64, AtomicU64, Ordering},
@@ -97,6 +98,12 @@ fn get_app_name() -> Option<String> {
     Some(basename.to_string_lossy().into_owned())
 }
 
+fn make_version() -> u32 {
+    pkg_version_major!() * 1000000
+        + pkg_version_minor!() * 1000
+        + pkg_version_patch!()
+}
+
 impl<C: Compositor> OpenXrData<C> {
     pub fn new(injector: &Injector) -> Result<Self, InitError> {
         #[cfg(not(test))]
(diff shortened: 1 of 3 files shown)


src/openxr_data.rs
```diff
@@ -97,6 +98,12 @@ fn get_app_name() -> Option<String> {
     Some(basename.to_string_lossy().into_owned())
 }
 
+fn make_version() -> u32 {
+    pkg_version_major!() * 1000000
+        + pkg_version_minor!() * 1000
+        + pkg_version_patch!()
```
Supreeeme: I don't think this dependency is needed, we can easily pull these numbers with `env!("CARGO_PKG_VERSION_MAJOR")` and friends - https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

xytovl:
> > The cargo version is currently 0.3.0, is there a way to get the number of commits since the tag?
> 
> This is probably best accomplished by setting an env var from the build script and using `git rev-list v0.3 --count` or something.
I wouldn't know how to integrate such a script, especially as it will have to work without git metadata. Distribution package usually pull a tarball, so there needs to be an override or a fallback.
I am leaving it as major/minor/patch for the moment.

Supreeeme:
lgtm

src/openxr_data.rs
```diff
@@ -97,6 +97,12 @@ fn get_app_name() -> Option<String> {
     Some(basename.to_string_lossy().into_owned())
 }
 
+fn make_version() -> u32 {
+    u32::from_str_radix(env!("CARGO_PKG_VERSION_MAJOR"), 10).unwrap_or(0) * 1000000
+        + u32::from_str_radix(env!("CARGO_PKG_VERSION_MINOR"), 10).unwrap_or(0) * 1000
+        + u32::from_str_radix(env!("CARGO_PKG_VERSION_PATCH"), 10).unwrap_or(1)
```
galister: `u32::from_str_radix(x, 10).unwrap_or(0)` → `x.parse::<u32>().unwrap_or(0)`

PR #214, opened 2025-10-09
Application name support for 32 wine

Mr-Zero88:


Supreeeme:
I'd rather not add that prefix to the application name, I prefer what #210 does.

galister:
https://github.com/Supreeeme/xrizer/pull/210 got merged, so I suggest just doing the wine-preloader part.

PR #222, opened 2025-11-04
src/overlay.rs: Implement IsOverlayVisible

mkopec:
With this change, The Talos Principle VR is now playable (via Proton, the native Linux port is still broken).


Supreeeme:
Thanks! Looks like it just needs `cargo fmt` run on it.

ImSapphire:
This is also required by Portal 2 VR

mkopec:
> Thanks! Looks like it just needs cargo fmt run on it.

Applied :)

PR #223, opened 2025-11-04
Refactor device logic (continued)

### LINKED ISSUE #57: No controllers detected when only right controller is on

ImSapphire:
Continuation of #68
I tried to update this and address the review comments, however there are some things I may have missed or accidentally broke during the merge. Original PR description below:

> This is the start of the device refactor work. I've decided to keep it as minimal as possible for now. This gives us enough of a skeleton to implement new devices that provide poses. Devices are now stored in a `Vec`, accessed via the `TrackedDeviceList` struct. They currently only provide a pose, and access to the device's interaction profile.

Passes all tests.

Part of #27
Fixes #57

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,6 @@
 [env]
 # TODO: get leak checker passing
-MIRIFLAGS = "-Zmiri-disable-isolation -Zmiri-permissive-provenance -Zmiri-ignore-leaks"
+MIRIFLAGS = "-Zmiri-disable-isolation -Zmiri-permissive-provenance -Zmiri-ignore-leaks -Zmiri-deterministic-floats"
 
 [alias]
 xbuild = "run --package xbuild --quiet --"


src/input/devices.rs
```diff
@@ -0,0 +1,427 @@
...
+    pub head: xr::Path,
+    pub left: xr::Path,
+    pub right: xr::Path,
+}
+
+impl SubactionPaths {
+    pub fn new(instance: &xr::Instance) -> Self {
+        let head = instance
+            .string_to_path("/user/head")
+            .expect("Failed to convert string to path");
```
Supreeeme: I'm not sure the head interaction profile is useful or necessary information.

src/input/devices.rs
```diff
@@ -0,0 +1,427 @@
...
+            .compare_exchange(!current, current, Ordering::Relaxed, Ordering::Relaxed)
+            .is_ok()
+    }
+
+    pub fn get_type(&self) -> TrackedDeviceType {
+        self.device_type
+    }
+
+    // Controllers
+    fn get_controller_pose(
```
Supreeeme: This could just be a standalone function, instead of a function on XrTrackedDevice

src/input/devices.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    pub fn get_controller_hand(&self) -> Option<Hand> {
+        match self.get_type() {
+            TrackedDeviceType::Controller { hand, .. } => Some(hand),
+            _ => None,
+        }
+    }
+
+    // HMD
+    fn get_hmd_pose(
```
Supreeeme: Same comment as get_controller_pose

src/input/devices.rs
```diff
@@ -0,0 +1,427 @@
...
+    Hmd,
+    Controller { hand: Hand },
+}
+
+impl From<Hand> for TrackedDeviceType {
+    fn from(hand: Hand) -> Self {
+        TrackedDeviceType::Controller { hand }
+    }
+}
+pub struct XrTrackedDevice {
```
```suggestion
pub struct TrackedDevice {
```

src/system.rs
```diff
@@ -668,18 +634,22 @@ impl vr::IVRSystem022_Interface for System {
...
+        self.input
+            .get()
+            .and_then(|input| match device_index {
+                x if input.device_index_to_hand(x).is_some() => input
+                    .get_controller_int_tracked_property(
+                        input.device_index_to_hand(x).unwrap(),
+                        prop,
+                    ),
+                _ => None,
+            })
```
Supreeeme:
```suggestion
            .and_then(|input| input.device_index_to_hand(device_index))
            .and_then(|hand| input.get_controller_int_tracked_property(hand, prop))
```

src/openxr_data.rs
```diff
@@ -646,59 +614,58 @@ impl SessionData {
 
 pub struct AtomicPath(AtomicU64);
 impl AtomicPath {
+    pub(crate) fn new() -> Self {
+        Self(0.into())
```
```suggestion
        Self(xr::Path::NULL.into())
```

src/openxr_data.rs
```diff
@@ -646,59 +614,58 @@ impl SessionData {
...
-
 #[repr(u32)]
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum Hand {
     Left = 1,
     Right,
 }
 
-impl TryFrom<vr::TrackedDeviceIndex_t> for Hand {
+pub type HandPath = &'static str;
```
Supreeeme: Don't love this type alias

src/input.rs
```diff
@@ -50,12 +53,13 @@ pub struct Input<C: openxr_data::Compositor> {
...
     set_map: RwLock<SlotMap<ActionSetKey, String>>,
     loaded_actions_path: OnceLock<PathBuf>,
-    cached_poses: Mutex<CachedSpaces>,
     legacy_state: legacy::LegacyState,
     skeletal_tracking_level: RwLock<vr::EVRSkeletalTrackingLevel>,
     profile_map: HashMap<xr::Path, &'static profiles::ProfileProperties>,
     estimated_finger_state: [Mutex<FingerState>; 2],
+    subaction_paths: SubactionPaths,
     events: Mutex<VecDeque<InputEvent>>,
+    devices: RwLock<TrackedDeviceList>,
```
Supreeeme, line 62: Putting TrackedDeviceList behind a RwLock feels bad to me, it seems a little prone to deadlocks. I feel like something like [ArcSwap](https://docs.rs/arc-swap/latest/arc_swap/type.ArcSwap.html) would feel better, though I'm not sure it's worth the effort.

src/input/devices.rs
```diff
@@ -0,0 +1,427 @@
...
+    device_type: TrackedDeviceType,
+    interaction_profile: Mutex<Option<&'static dyn InteractionProfile>>,
+    profile_path: AtomicPath,
+    connected: AtomicBool,
+    previous_connected: AtomicBool,
+    pose_cache: Mutex<Option<vr::TrackedDevicePose_t>>,
```
Supreeeme: Considering that:
1. TrackedDevices are behind the TrackedDeviceList, which is behind a RwLock (currently)
2. Most of these only ever change when a RwLock is taken on the TrackedDeviceList anyway

I don't see why these need to be atomic/Mutexes.

src/input/devices.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    pub fn get_profile_path(&self) -> xr::Path {
+        self.profile_path.load()
+    }
+
+    pub fn set_profile_path(&self, path: xr::Path) {
+        self.profile_path.store(path);
+    }
+
```
Supreeeme: If the atomics are removed, I don't think these getters/setters are necessary, can just make the fields pub.

src/input/devices.rs
ImSapphire: It could be used to allow usage of the buttons on the `/interaction_profiles/htc/vive_pro` interaction profile, but if this isn't cared about I can remove that code

src/input/devices.rs
ImSapphire: Hm, if I make `previous_connected` a bool, I'd have to take a write lock in `Input::get_next_event` so `previous_connected` can be mutated in `has_connected_changed`, not sure if that's ideal or how I can solve that...

src/input.rs
```diff
@@ -823,12 +840,20 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
...
+                    bActive: true,
+                    activeOrigin: active_origin,
+                    pose,
+                });
+            } else {
+                action_data.write(vr::InputPoseActionData_t {
+                    bActive: false,
+                    ..Default::default()
+                });
+            }
```
```suggestion
            let pose = self.get_controller_pose(hand, Some(origin)).unwrap_or_default()
            action_data.write(vr::InputPoseActionData_t {
                    bActive: true,
                    activeOrigin: active_origin,
                    pose,
              });
```

src/input.rs
```diff
@@ -739,28 +752,32 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
...
-                    x if x == self.openxr.right_hand.subaction_path => (
-                        Some(Hand::Right),
-                        Some(self.openxr.right_hand.profile_path.load()),
-                    ),
+                    x if x == self.get_subaction_path(Hand::Left) => left_hand
+                        .map(|h| (Hand::Left, h.profile_path))
+                        .unzip(),
+                    x if x == self.get_subaction_path(Hand::Right) => right_hand
+                        .map(|h| (Hand::Right, h.profile_path))
+                        .unzip(),
```
Supreeeme: Maybe:
```rust
let devices = self.devices.read().unwrap();
let get_hand = |hand| devices.get_controller(hand).map(|h| hand, h.profile_path).unzip();
let (active_origin, hand) = match loaded.try_get_action(action) {
            Ok(ActionData::Pose) => {
                let (mut hand, interaction_profile) = match subaction_path {
                    x if x == self.get_subaction_path(Hand::Left) => get_hand(Hand::Left),
                    x if x == self.get_subaction_path(Hand::Right) => get_hand(Hand::Right)
```

src/input.rs
```diff
@@ -1060,8 +1090,12 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
...
-            unsync_custom_bindings(*key, left_profile);
-            if left_profile != right_profile {
+            if let Some(left_profile) = left_profile {
+                unsync_custom_bindings(*key, left_profile);
+                if right_profile.is_some_and(|p| p != left_profile) {
+                    unsync_custom_bindings(*key, right_profile.unwrap());
+                }
+            } else if let Some(right_profile) = right_profile {
                 unsync_custom_bindings(*key, right_profile);
             }
```
```suggestion
            match (left_profile, right_profile) {
                (Some(profile), None) | (None, Some(profile)) => unsync_custom_bindings(*key, profile),
                (Some(left), Some(right)) => {
                    unsync_custom_bindings(*key, left);
                    if left != right {
                        unsync_custom_bindings(*key, right);
                    }
                }
            }
```

src/input/devices.rs
```diff
@@ -0,0 +1,386 @@
...
+        let left = instance
+            .string_to_path("/user/hand/left")
+            .expect("Failed to convert string to path");
+        let right = instance
+            .string_to_path("/user/hand/right")
+            .expect("Failed to convert string to path");
+
+        Self { left, right }
+    }
+}
```
```suggestion
pub struct SubactionPaths {
    pub left: xr::Path,
    pub right: xr::Path,
}

impl SubactionPaths {
    pub fn new(instance: &xr::Instance) -> Self {
        let left = instance
            .string_to_path("/user/hand/left")
            .expect("Failed to convert string to path");
        let right = instance
            .string_to_path("/user/hand/right")
            .expect("Failed to convert string to path");

        Self { left, right }
    }
}

pub struct TrackedDeviceList {
    devices: Vec<TrackedDevice>,
}
```

Far away from its impl :)

src/input/devices.rs
```diff
@@ -0,0 +1,386 @@
...
+    pub(super) fn get_controller_mut(&mut self, hand: Hand) -> Option<&mut TrackedDevice> {
+        self.get_device_mut(self.get_controller_index(hand))
+    }
+
+    fn get_controller_index(&self, hand: Hand) -> vr::TrackedDeviceIndex_t {
+        self.iter()
+            .enumerate()
+            .find(|(_, device)| device.get_controller_hand() == Some(hand))
+            .map(|(i, _)| i as vr::TrackedDeviceIndex_t)
+            .unwrap_or(vr::k_unTrackedDeviceIndexInvalid)
```
```suggestion
    fn get_controller_index(&self, hand: Hand) -> Option<vr::TrackedDeviceIndex_t> {
        self.iter()
            .enumerate()
            .find(|(_, device)| device.get_controller_hand() == Some(hand))
            .map(|(i, _)| i as vr::TrackedDeviceIndex_t)
```

src/input/devices.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    pub fn iter(&self) -> std::slice::Iter<'_, TrackedDevice> {
+        self.devices.iter()
+    }
+    pub fn iter_mut(&mut self) -> std::slice::IterMut<'_, TrackedDevice> {
```
```suggestion
    pub fn iter(&self) -> impl Iterator<Item = &TrackedDevice> {
        self.devices.iter()
    }
    pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut TrackedDevice> {
```

src/system.rs
```diff
@@ -596,9 +556,11 @@ impl vr::IVRSystem023_Interface for System {
...
             },
-            x if Hand::try_from(x).is_ok() => self.input.get().and_then(|i| {
-                i.get_controller_string_tracked_property(Hand::try_from(x).unwrap(), prop)
-            }),
+            x if input.device_index_to_hand(x).is_some() => input
+                .get_controller_string_tracked_property(
+                    input.device_index_to_hand(x).unwrap(),
+                    prop,
+                ),
             _ => None,
```
Supreeeme:
```suggestion
            x => input
                .device_index_to_hand(x).map(|x| input.get_controller_string_tracked_property(x, prop)),
```
```

src/input.rs
ImSapphire: Had no idea that first expression was possible, nice!

Supreeeme (approved):
I'm still sus on the RwLock but I don't really have a better answer right now, and it can be changed later. I'll merge after I fix the CI with #227.

PR #229, opened 2025-11-13
Integrate `libmonado` and use it to pass battery status

### LINKED ISSUE #26: Add optional libmonado data handler as supplement
Libmonado can be used in a more experimental and speculative manner to allow access to missing OpenVR properties and is provided by the libmonado-rs crate https://github.com/technobaboo/libmonado-rs

Missing mappings of OpenVR->OpenXR can be directed here when available.

jvyden:
Closes #26.

# Summary

This PR integrates the `libmonado` library through the `libmonado-rs` crate, the same crate `wlx-overlay-s` and others use. I've integrated it as an optional (but enabled by default) feature, with the ability for it to be either entirely compiled out or be silently unavailable on startup if we can't find Monado.

As a demonstration of usage of the new library, I've also integrated the `DeviceIsCharging_Bool`, `DeviceProvidesBatteryStatus_Bool`, and `DeviceBatteryPercentage_Float` tracked properties to allow games to observe device battery status.

# Testing

I've tested this on WiVRn and it seems to work well for headset battery. I do not have a headset that reports controller battery through OpenXR, so I had a friend test it on their Index and it also worked.

In both cases, I've tested in Resonite using the game's built-in scripting language to observe the values reported directly.

Additionally, on WiVRn you need to specify `LIBMONADO_PATH` before xrizer can connect to Monado from within Proton's container. This seems to work fine on Monado for a reason I'm not sure why, but I've made an issue to track a workaround in `libmonado-rs`: https://github.com/technobaboo/libmonado-rs/issues/24

# Notes

- This is my first time integrating a feature like this into a Rust project, and I'm not sure if I'm initializing `libmonado` in the correct place. Guidance here and with code quality would be appreciated as I'm not familiar with Rust.
- I'm not *entirely certain* that `libmonado` is thread-safe, but over the past 3-4 days I've had no issues in release builds with wrapping the library context in a `Send`/`Sync`.
- Unsure how to integrate unit tests into this. Should I not care about this since the bulk of these features are in other projects?

### DIFF
--- a/src/monado.rs
+++ b/src/monado.rs
@@ -25,13 +25,13 @@ impl SafeMonado {
     }
 
     pub fn get_device_from_vr_index(&self, device_index: u32) -> Option<Device<'_>> {
-        return match device_index {
+        match device_index {
             vr::k_unTrackedDeviceIndex_Hmd => self.0.device_from_role(DeviceRole::Head).ok(),
             x if Hand::try_from(x).is_ok() => match Hand::try_from(x).unwrap() {
                 Hand::Left => self.0.device_from_role(DeviceRole::Left).ok(),
                 Hand::Right => self.0.device_from_role(DeviceRole::Right).ok(),
             },
             _ => None,
-        };
+        }
     }
 }
--- a/src/system.rs
+++ b/src/system.rs
@@ -211,16 +211,12 @@ impl System {
     #[cfg(feature = "monado")]
     pub fn mnd_get_device_battery(&self, device_index: u32) -> Option<BatteryStatus> {
         let session_data = self.openxr.session_data.get();
-        let monado = match &session_data.monado {
-            Some(m) => m,
-            None => return None,
-        };
 
-        let status = monado
+        session_data
+            .monado
+            .as_ref()?
             .get_device_from_vr_index(device_index)
-            .and_then(|device| device.battery_status().ok());
-
-        return status;
+            .and_then(|device| device.battery_status().ok())
     }
 }
 
@@ -754,7 +750,7 @@ impl vr::IVRSystem023_Interface for System {
             vr::ETrackedDeviceProperty::DeviceBatteryPercentage_Float => {
                 let Some(d) = self.mnd_get_device_battery(device_index) else {
                     if let Some(err) = unsafe { err.as_mut() } {
-                        *err = vr::ETrackedPropertyError::UnknownProperty;
+                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
                     }
                     return 0.0;
                 };
@@ -778,12 +774,7 @@ impl vr::IVRSystem023_Interface for System {
                     return views[1].pose.position.x - views[0].pose.position.x;
                 }
                 vr::ETrackedDeviceProperty::DisplayFrequency_Float => return 90.0,
-                _ => {
-                    if let Some(err) = unsafe { err.as_mut() } {
-                        *err = vr::ETrackedPropertyError::UnknownProperty;
-                    }
-                    return 0.0;
-                }
+                _ => {}
             }
         }
 
@@ -806,12 +797,16 @@ impl vr::IVRSystem023_Interface for System {
             return false;
         }
 
+        if let Some(err) = unsafe { err.as_mut() } {
+            *err = vr::ETrackedPropertyError::Success;
+        }
+
         match prop {
             #[cfg(feature = "monado")]
             vr::ETrackedDeviceProperty::DeviceProvidesBatteryStatus_Bool => {
                 let Some(d) = self.mnd_get_device_battery(device_index) else {
                     if let Some(err) = unsafe { err.as_mut() } {
-                        *err = vr::ETrackedPropertyError::UnknownProperty;
+                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
                     }
                     return false;
                 };
@@ -822,7 +817,7 @@ impl vr::IVRSystem023_Interface for System {
             vr::ETrackedDeviceProperty::DeviceIsCharging_Bool => {
                 let Some(d) = self.mnd_get_device_battery(device_index) else {
                     if let Some(err) = unsafe { err.as_mut() } {
-                        *err = vr::ETrackedPropertyError::UnknownProperty;
+                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
                     }
                     return false;
                 };


src/monado.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    pub fn get_device_from_vr_index(&self, device_index: u32) -> Option<Device<'_>> {
+        return match device_index {
+            vr::k_unTrackedDeviceIndex_Hmd => self.0.device_from_role(DeviceRole::Head).ok(),
+            x if Hand::try_from(x).is_ok() => match Hand::try_from(x).unwrap() {
+                Hand::Left => self.0.device_from_role(DeviceRole::Left).ok(),
+                Hand::Right => self.0.device_from_role(DeviceRole::Right).ok(),
+            },
+            _ => None,
+        };
```
ImSapphire:
```suggestion
        match device_index {
            vr::k_unTrackedDeviceIndex_Hmd => self.0.device_from_role(DeviceRole::Head).ok(),
            x if Hand::try_from(x).is_ok() => match Hand::try_from(x).unwrap() {
                Hand::Left => self.0.device_from_role(DeviceRole::Left).ok(),
                Hand::Right => self.0.device_from_role(DeviceRole::Right).ok(),
            },
            _ => None,
        }
```

src/system.rs
```diff
@@ -203,6 +207,21 @@ impl System {
...
+        let monado = match &session_data.monado {
+            Some(m) => m,
+            None => return None,
+        };
+
+        let status = monado
+            .get_device_from_vr_index(device_index)
+            .and_then(|device| device.battery_status().ok());
+
+        return status;
```
ImSapphire:
```suggestion
        session_data
            .monado
            .as_ref()?
            .get_device_from_vr_index(device_index)
            .and_then(|device| device.battery_status().ok())
```

ImSapphire: The `.as_ref()` might not be needed here, not sure

src/system.rs
```diff
@@ -716,29 +735,62 @@ impl vr::IVRSystem023_Interface for System {
...
+                vr::ETrackedDeviceProperty::UserIpdMeters_Float => {
+                    let views = self.get_views(xr::ReferenceSpaceType::VIEW).views;
+                    return views[1].pose.position.x - views[0].pose.position.x;
+                }
+                vr::ETrackedDeviceProperty::DisplayFrequency_Float => return 90.0,
+                _ => {
+                    if let Some(err) = unsafe { err.as_mut() } {
+                        *err = vr::ETrackedPropertyError::UnknownProperty;
+                    }
+                    return 0.0;
```
ImSapphire: duplicated below, this block can be left empty

src/system.rs
```diff
@@ -747,6 +799,46 @@ impl vr::IVRSystem023_Interface for System {
         err: *mut vr::ETrackedPropertyError,
     ) -> bool {
         debug!(target: log_tags::TRACKED_PROP, "requesting bool property: {prop:?} ({device_index})");
+        if !self.IsTrackedDeviceConnected(device_index) {
+            if let Some(err) = unsafe { err.as_mut() } {
+                *err = vr::ETrackedPropertyError::InvalidDevice;
+            }
+            return false;
+        }
+
```
ImSapphire, line 799:
```suggestion

    if let Some(err) = unsafe { err.as_mut() } {
        *err = vr::ETrackedPropertyError::Success;
    }

```

src/system.rs
```diff
@@ -747,6 +799,46 @@ impl vr::IVRSystem023_Interface for System {
...
+                    return false;
+                };
+
+                return d.present;
+            }
+            #[cfg(feature = "monado")]
+            vr::ETrackedDeviceProperty::DeviceIsCharging_Bool => {
+                let Some(d) = self.mnd_get_device_battery(device_index) else {
+                    if let Some(err) = unsafe { err.as_mut() } {
+                        *err = vr::ETrackedPropertyError::UnknownProperty;
```
ImSapphire:
```suggestion
                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
```

src/system.rs
```diff
@@ -716,29 +735,62 @@ impl vr::IVRSystem023_Interface for System {
...
+
         match prop {
-            vr::ETrackedDeviceProperty::UserIpdMeters_Float => {
-                let views = self.get_views(xr::ReferenceSpaceType::VIEW).views;
-                views[1].pose.position.x - views[0].pose.position.x
+            #[cfg(feature = "monado")]
+            vr::ETrackedDeviceProperty::DeviceBatteryPercentage_Float => {
+                let Some(d) = self.mnd_get_device_battery(device_index) else {
+                    if let Some(err) = unsafe { err.as_mut() } {
+                        *err = vr::ETrackedPropertyError::UnknownProperty;
```
ImSapphire:
```suggestion
                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
```

src/system.rs
```diff
@@ -747,6 +799,46 @@ impl vr::IVRSystem023_Interface for System {
...
+            return false;
+        }
+
+        match prop {
+            #[cfg(feature = "monado")]
+            vr::ETrackedDeviceProperty::DeviceProvidesBatteryStatus_Bool => {
+                let Some(d) = self.mnd_get_device_battery(device_index) else {
+                    if let Some(err) = unsafe { err.as_mut() } {
+                        *err = vr::ETrackedPropertyError::UnknownProperty;
+                    }
```
ImSapphire, line 810:
```suggestion
```

jvyden: i think what we actually want here is `ValueNotProvidedByDevice`?

jvyden: did so in 820fc0e but lmk if you want still want UnknownProperty by default

src/system.rs
```diff
@@ -716,29 +731,57 @@ impl vr::IVRSystem023_Interface for System {
...
+                };
+
+                if !d.present {
+                    if let Some(err) = unsafe { err.as_mut() } {
+                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
+                    }
+                    return 0.0;
+                }
+
+                return d.charge;
```
Supreeeme, line 765:
```suggestion
                let battery = self.mnd_get_device_battery(device_index);
                if let Some(charge) = battery.and_then(|b| b.present.then_some(b.charge)) {
                    return charge;
                } else {
                    if let Some(err) = unsafe { err.as_mut() } {
                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
                    }
                    return 0.0;
                }
```

src/system.rs
```diff
@@ -747,6 +790,50 @@ impl vr::IVRSystem023_Interface for System {
...
+            #[cfg(feature = "monado")]
+            vr::ETrackedDeviceProperty::DeviceProvidesBatteryStatus_Bool => {
+                let Some(d) = self.mnd_get_device_battery(device_index) else {
+                    if let Some(err) = unsafe { err.as_mut() } {
+                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
+                    }
+                    return false;
+                };
+
+                return d.present;
```
Supreeeme, line 814:
```suggestion
                if let Some(d) = self.mnd_get_device_battery(device_index) {
                    return d.present;
                } else {
                    if let Some(err) = unsafe { err.as_mut() } {
                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
                    }
                    return false;
                };
```

src/system.rs
```diff
@@ -747,6 +790,50 @@ impl vr::IVRSystem023_Interface for System {
...
+                };
+
+                if !d.present {
+                    if let Some(err) = unsafe { err.as_mut() } {
+                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
+                    }
+                    return false;
+                }
+
+                return d.charging;
```
Supreeeme, line 832:
```suggestion
                let battery = self.mnd_get_device_battery(device_index);
                if let Some(charging) = battery.and_then(|b| b.present.then_some(b.charging)) {
                    return charging;
                } else {
                    if let Some(err) = unsafe { err.as_mut() } {
                        *err = vr::ETrackedPropertyError::ValueNotProvidedByDevice;
                    }
                    return false;
                }
```

src/system.rs
```diff
@@ -203,6 +207,17 @@ impl System {
         let mut views = self.views.lock().unwrap();
         views.get_views(&session, self.openxr.display_time.get(), ty)
     }
+
+    #[cfg(feature = "monado")]
+    pub fn mnd_get_device_battery(&self, device_index: u32) -> Option<BatteryStatus> {
```
Supreeeme, line 212:
```suggestion
    fn mnd_get_device_battery(&self, device_index: u32) -> Option<BatteryStatus> {
```

src/monado.rs
```diff
@@ -0,0 +1,37 @@
...
+use log::{info, warn};
+
+use crate::openxr_data::Hand;
+
+use openvr as vr;
+
+pub struct SafeMonado(pub Monado);
+
+unsafe impl Send for SafeMonado {}
+unsafe impl Sync for SafeMonado {}
```
Supreeeme, line 11: This needs a comment explaining why it's done

src/system.rs
ImSapphire: it probably doesn't matter in practice, but I don't think returning an error when the battery isn't available makes sense for this property, unsure what SteamVR does

galister:
It's important to know that libmonado uses a separate IPC session to communicate with Monado, essentially the same as what libopenxr_monado does.

There are 2 major issues with this:
- Monado has a low-low IPC client limit of 8. Once the limit is hit, it's not possible to launch more XR client applications. I have already ran into issues with just tools like Motoc and Wlx, and have ended up bumping the limit on my local build to 16.
- ~~While libmonado's IPC session counts as a headless session, there's a shortcoming in Monado that makes it create a full compositor object even for headless sessions. Due to this, these IPC sessions may not be so lightweight as they seem.~~ _Edit: this only happens with MND_HEADLESS, not with libmonado._

There is an unpublished extension `EXT_battery_state_display`, currently offered only by SteamVR. I would personally consider waiting for that to be finalized, as that would allow us to have a less intrusive solution.

ImSapphire:
> Monado has a low-low IPC client limit of 8. Once the limit is hit, it's not possible to launch more XR client applications. I have already ran into issues with just tools like Motoc and Wlx, and have ended up bumping the limit on my local build to 16.

Maybe we should have this bumped upstream?

> ~While libmonado's IPC session counts as a headless session, there's a shortcoming in Monado that makes it create a full compositor object even for headless sessions. Due to this, these IPC sessions may not be so lightweight as they seem.~ _Edit: this only happens with MND_HEADLESS, not with libmonado._

This should be fixed in https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2606

galister:
Monado IPC client increase: https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2644

SpookySkeletons:
> Monado IPC client increase: https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2644

Now merged.

SpookySkeletons:
Mind giving this a rebase?

jvyden:
Sure, sorry, forgot about this branch. I should have this up to date tomorrow.

PR #230, opened 2025-11-17
openxr_data: fire VREvent_Quit event when session state changes to STOPPING

ImSapphire:
Tested in VRChat via closing with WiVRn's quit button.

### DIFF
--- a/src/openxr_data.rs
+++ b/src/openxr_data.rs
@@ -1,10 +1,11 @@
 use crate::{
     clientcore::{Injected, Injector},
     graphics_backends::{GraphicsBackend, VulkanData, supported_apis_enum},
+    system::System,
 };
 use derive_more::Deref;
 use glam::f32::{Quat, Vec3};
-use log::{info, warn};
+use log::{debug, info, warn};
 use openvr as vr;
 use openxr as xr;
 use std::mem::ManuallyDrop;
@@ -47,6 +48,7 @@ pub struct OpenXrData<C: Compositor> {
     /// should only be externally accessed for testing
     pub(crate) input: Injected<crate::input::Input<C>>,
     pub(crate) compositor: Injected<C>,
+    pub(crate) system: Injected<System>,
 }
 
 impl<C: Compositor> Drop for OpenXrData<C> {
@@ -191,6 +193,7 @@ impl<C: Compositor> OpenXrData<C> {
             enabled_extensions: exts,
             input: injector.inject(),
             compositor: injector.inject(),
+            system: injector.inject(),
         })
     }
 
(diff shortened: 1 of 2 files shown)


src/system.rs
```diff
@@ -483,6 +489,13 @@ impl vr::IVRSystem023_Interface for System {
         size: u32,
         pose: *mut vr::TrackedDevicePose_t,
     ) -> bool {
+        if let Some(event_) = self.events.lock().unwrap().pop_front() {
+            unsafe {
+                event.write(event_);
+            }
```
Supreeeme, line 547: I would be careful about this. The size and layout of `vr::VREvent_t` changes across OpenVR versions. For example, the `bConnectionLost` field doesn't exist until 1.3.22. It might be good to take some inspiration from what `Input::get_next_event` does in terms of checking sizes and the layout.

src/openxr_data.rs
```diff
@@ -179,6 +183,25 @@ impl<C: Compositor> OpenXrData<C> {
                 xr::Event::SessionStateChanged(event) => {
                     state = Some(event.state());
                     info!("OpenXR session state changed: {:?}", event.state());
+                    if let Some(system) = (state == Some(xr::SessionState::LOSS_PENDING))
+                        .then_some(self.system.get())
+                        .flatten()
```
```suggestion
                    if let Some(system) = self.system.get().filter(|_| state == Some(xr::SessionState::LOSS_PENDING))
```

Supreeeme:
Sorry for the delay. Looks good, but a test for this would be cool.

ImSapphire:
> Sorry for the delay. Looks good, but a test for this would be cool.

I'm not exactly sure how to write unit tests, and I'm working on some other things right now, would it be fine to merge without a test?

Supreeeme:
1. I realized this is actually session loss, not instance loss. The commit message should probably be reworded.
2. I was thinking about this and came to a realization: is telling the application on session loss really the right thing here? To me `VREvent_Quit` seems more like a way to tell an application to gracefully close, but session loss seems closer to an error condition, and if we have applications gracefully exit it won't be obvious something erroneous happened. I would think session state being STOPPING or EXITING would be more appropriate.

ImSapphire:
@Supreeeme Would emitting `VREvent_Quit` event on change to `STOPPING` state be enough? I thought there would be more involved in that case, though I guess nothing in OpenXR mandates that the app stop immediately

Supreeeme:
Based on the spec, I think the correct thing to do would be:
1. Send VREvent_Quit on STOPPING
2. Call `end_session` (`xrEndSession`) when the application calls `AcknowledgeQuit_Exiting`

ImSapphire:
iirc in SteamVR, AcknowledgeQuit_Exiting just extends SteamVR's kill timer to give the app time to safely shutdown, if we stop the session there we'd probably crash because the game would keep calling functions like WaitGetPoses, Get*ActionData etc. Not sure what the right thing to do there is

ImSapphire:
Some apps also simply don't call AcknowledgeQuit_Exiting (VRChat, Resonite from what I've tested so far)

ImSapphire:
Added a test

ImSapphire:
@Supreeeme is there anything I could do to get this merged?

src/openxr_data.rs
```diff
@@ -206,6 +209,25 @@ impl<C: Compositor> OpenXrData<C> {
                 xr::Event::SessionStateChanged(event) => {
                     state = Some(event.state());
                     info!("OpenXR session state changed: {:?}", event.state());
+                    if let Some(system) = self.system.get().filter(|_| {
+                        (cfg!(test) || session_data.temp_vulkan.is_none())
```
Supreeeme, line 217: Why is this unconditional for tests?

ImSapphire: When running the test, the session is using a temp vulkan instance, that seemed like the easiest way to get it working to me

Supreeeme: I don't love adding test specific logic to actual code. I think it would be better to have some kind of `use_real_session` helper method in the test that handles doing this.

PR #233, opened 2025-11-20
dont die on unexpected pose type

galister:
I discovered this in the case of ValheimVR (mod) which asks for `base` on `vive_controller` and `touch_controller`, as well as `handgrip` on `index_controller`. When running on `main` branch, bindings would fail to load and thus not provide any input whatsoever. However, the same title works fine with OpenComposite.

The [docs](https://github.com/ValveSoftware/openvr/wiki/Input-Profiles#pose-components) hint at the fact that Raw and Tip are merely the ones that are _always_ provided, with ability to add custom ones using external files.

Since we currently don't consider such files, the second best option would be to fall back to a sane default. A warning is left in place in order to help identify issues that stem from the fallback not being a good fit.

### DIFF
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -1000,8 +1000,13 @@ impl vr::IVROverlay028_Interface for OverlayMan {
     ) -> vr::EVROverlayError {
         todo!()
     }
-    fn IsOverlayVisible(&self, _: vr::VROverlayHandle_t) -> bool {
-        todo!()
+    fn IsOverlayVisible(&self, handle: vr::VROverlayHandle_t) -> bool {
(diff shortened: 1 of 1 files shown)


Mr-Zero88:
this is a WIP extraction from the hl2vr-fixes branch:
https://github.com/Supreeeme/xrizer/pull/234

Supreeeme:
I guess potentially incorrect hand positions is better than no hand positions/input at all? And some of the pose types might just end up being the same as `raw`. On the other hand this could make it more challenging to determine why a pose position is offset. I guess seeing the logs might just be enough though.

Given this, I guess I'll accept this over #234.

Issue #224, opened 2025-11-06
[Bug]: Township Tale - no controllers in game "Failed to parse bindings"


Mhowser:
### Operating System

Arch Linux

### OpenXR Runtime

Wivrn

### Describe the issue.

Xrizer log:
```
[2025-11-05T21:15:07.660 INFO  xrizer ThreadId(1)] Initializing XRizer version v0.3-35-gc77db6c
[2025-11-05T21:15:07.660 INFO  xrizer::clientcore ThreadId(1)] Creating ClientCore version "IVRClientCore_003"
[2025-11-05T21:15:07.696 INFO  xrizer::openxr_data ThreadId(1)] Creating OpenXR session with graphics API openxr::graphics::vulkan::Vulkan
[2025-11-05T21:15:07.696 INFO  xrizer::openxr_data ThreadId(1)] New session created!
[2025-11-05T21:15:07.696 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: READY
[2025-11-05T21:15:07.696 INFO  xrizer::openxr_data ThreadId(1)] Began OpenXR session.
[2025-11-05T21:15:07.697 INFO  xrizer::input ThreadId(1)] loading action manifest from "/home/kit/Documents/township tale/A Township Tale/A Township Tale_Data/StreamingAssets/SteamVR/actions.json"
[2025-11-05T21:15:07.697 ERROR xrizer::input::action_manifest ThreadId(1)] Failed to parse bindings for ViveController: unknown variant `handgrip`, expected one of `raw`, `tip`, `gdc2015` at line 22 column 13
[2025-11-05T21:15:07.697 INFO  xrizer::input::action_manifest ThreadId(1)] Ignoring bindings for unknown profile vive_cosmos_controller
[2025-11-05T21:15:07.697 ERROR xrizer::input::action_manifest ThreadId(1)] Failed to parse bindings for OculusTouch: unknown variant `handgrip`, expected one of `raw`, `tip`, `gdc2015` at line 22 column 13
[2025-11-05T21:15:07.697 ERROR xrizer::input::action_manifest ThreadId(1)] Failed to parse bindings for Knuckles: unknown variant `handgrip`, expected one of `raw`, `tip`, `gdc2015` at line 22 column 13
[2025-11-05T21:15:07.697 INFO  xrizer::input::action_manifest ThreadId(1)] Ignoring bindings for unknown profile holographic_controller
[...]

Issue #213, opened 2025-10-08
[Bug]: Isuses with Quest2 controllers in HL2VR


Damianu:
### Operating System

Arch Linux

### OpenXR Runtime

WiVRN

### Describe the issue.

Grab doesn't work when using xrizer with quest 2 controllers.
It's normally bound to the grip button.

Additionally you can jump/crouch only when capacitive button on trigger is held, otherwise the actions are ignored.

Packages used (AUR):
- xrizer-multilib-git 1990df8-1
- wivrn-dashboard 25.9-1
- wivrn-multilib-git 25.9.64.gc6e05754-1
[Link to the logs](https://paste.cachyos.org/p/d65de37.txt)

### Steps to reproduce

Launch hl2vr, popup will say "Grab UNBOUND" and grab won't work.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Lamplions123:
Having the same issue, also arch with these packages

Bloxxel64:
same here, makes the game a little annoying

Issue #172, opened 2025-08-21
[Bug]: Valheim VR (VHVR) with Pico 4 and WiVRn


Vixenin:
### Operating System

Nixos

### OpenXR Runtime

WiVRn

### Describe the issue.

In VHVR which is a VR mode for Valheim the controllers for my Pico 4 are not being detected while using xrizer, but open composite can detect them and even they also have their own issues.

Link to the VR mode:
https://www.nexusmods.com/valheim/mods/847

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Supreeeme:
Can you launch the game and post the log here? The log is in `~/.local/state/xrizer/xrizer.txt`

Vixenin:
```
[2025-08-21T18:34:40.043 INFO  xrizer ThreadId(1)] Initializing XRizer
[2025-08-21T18:34:40.043 INFO  xrizer::clientcore ThreadId(1)] Creating ClientCore version "IVRClientCore_003"
[2025-08-21T18:34:40.096 INFO  xrizer::openxr_data ThreadId(1)] New session created!
[2025-08-21T18:34:40.096 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: READY
[2025-08-21T18:34:40.096 INFO  xrizer::openxr_data ThreadId(1)] Began OpenXR session.
[2025-08-21T18:34:40.096 INFO  xrizer::input ThreadId(1)] loading action manifest from "/home/vixenin/.local/share/Steam/steamapps/common/Valheim/valheim_Data/StreamingAssets/SteamVR/actions.json"
[2025-08-21T18:34:40.096 ERROR xrizer::input::action_manifest ThreadId(1)] Failed to parse bindings for ViveController: unknown variant `base`, expected `raw` or `gdc2015` at line 279 column 13
[2025-08-21T18:34:40.096 ERROR xrizer::input::action_manifest ThreadId(1)] Failed to parse bindings for OculusTouch: unknown variant `base`, expected `raw` or `gdc2015` at line 232 column 13
[2025-08-21T18:34:40.096 ERROR xrizer::input::action_manifest ThreadId(1)] Failed to parse bindings for Knuckles: unknown variant `handgrip`, expected `raw` or `gdc2015` at line 117 column 13
[2025-08-21T18:34:40.096 INFO  xrizer::input::action_manifest ThreadId(1)] Ignoring bindings for unknown profile holographic_controller
[...]

Issue #149, opened 2025-07-25
[Bug]: Controllers do not work properly in The Walking Dead: Saints & Sinners; Stuck in Menus


Alex52Github:
### Operating System

Linux Mint 22.1 Cinnamon

### OpenXR Runtime

WiVRn

### Describe the issue.

The walking dead Saints & Sinners does boot on xrizer.
The issue is that controllers don't seem to work properly. There's a prompt to reset your HMD view (which is usually done by clicking in right stick); I can get past it by pushing the R key on the keyboard, but i cant use controllers in the menu. I did notice that if I hold down my left menu button I can pull and drag my position in the menu scene, but that's all I can do

### Steps to reproduce

Open the game using xrizer, push R key

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

Supreeeme:
Can you post the log? `~/.local/state/xrizer/xrizer.txt`

Alex52Github:
> Can you post the log? `~/.local/state/xrizer/xrizer.txt`

Yes, do you see the issue? Is it an easy fix?

[xrizer.txt](https://github.com/user-attachments/files/21441046/xrizer.txt)

snrubthinks:
Seems like this is a game bug primarily with Quest headsets plus streaming where the headset and controllers are detected as an older Vive system. Virtual desktop users and the like seem to rely on SteamVR community bindings to fix it, but I'm not sure how that is handled here on a WiVRn + xrizer setup.

snrubthinks:
A similar issue occurs with The Golf Club VR. In this case, OpenComposite does not have the issue (it detects an Oculus touch style controller and input works correctly).

snrubthinks:
With TWDSS, you can get a bit further by  renaming  `oculus_touch_TWD.json` to `oculus_touch.json` in `TWD/Config/SteamVRBindings` (or using the xrizer binding override system). You may also need to set `bEnableOculusSteamVR=True` in `drive_c/users/steamuser/AppData/Local/TWD/Saved/Config/WindowsNoEditor/GameUserSettings.ini` of the game's Proton prefix.
 
Ironically, it works not because the new bindings "work", but because looking at the logs, it now falls back to seeing an `index_controller` instead of `touch_controller`. This allows pressing the two thumb sticks in to get past the menu screen. Controller pointers work, and hand/controller movements work, but movement of the character, grabbing weapons etc doesn't work.

[...]

PR #234, opened 2025-11-20
Support all the BoundPoseTypes [HL2VR / Township Tale / Valheim VR]

Mr-Zero88:
feat: Add new BoundPoseType variants, update pose parsing logic and warning for unsupported non-raw pose types

fixes:
- https://github.com/Supreeeme/xrizer/issues/224
- https://github.com/Supreeeme/xrizer/issues/172
- https://github.com/Supreeeme/xrizer/issues/149
- https://github.com/Supreeeme/xrizer/issues/213 (indirect)


Mr-Zero88:
Since there are more poses that can be defined on a per-controller basis for different controllers. Maybe instead of the type, we should just add a pose offset property that gets filled on load and raw just is an empty offset?

PR #239, opened 2025-11-26
Refactor VRClientCoreFactory to return only one ClientCore instance and OpenXR initialization to skip if already initialized (Portal2VR)

Mr-Zero88:
Return only one ClientCore instance from VRClientCoreFactory

Refactor OpenXR initialization to skip if already initialized

### DIFF
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -944,7 +944,8 @@ impl vr::IVROverlay028_Interface for OverlayMan {
         _: *const vr::VROverlayIntersectionParams_t,
         _: *mut vr::VROverlayIntersectionResults_t,
     ) -> bool {
-        todo!()
+        crate::warn_unimplemented!("ComputeOverlayIntersection");
+        false
     }
     fn SetOverlayMouseScale(
         &self,


src/lib.rs
```diff
@@ -217,8 +225,16 @@ pub unsafe extern "C" fn VRClientCoreFactory(
...
+        });
+
+        if let Some(c) = ret {
+            C.set(ClientCorePtr(c)).unwrap_or_else(|_| unreachable!());
+        } else {
+            return std::ptr::null_mut();
+        }
+    }
+
+    C.get().unwrap().0
```
Supreeeme:
```suggestion
        if let Some(c) = ret {
            C.set(ClientCorePtr(c)).unwrap_or_else(|_| unreachable!());
        }
    }

    C.get().map(|c| c.0).unwrap_or(std::ptr::null_mut())
```

ada-tv:
(after messing around to get 32 bit working) works for getting in-game on portal 2 vr with the monado qwerty driver :+1:

image

image
(weird behavior with the menu overlay, i might look into it after this pr is merged)

Mr-Zero88:
i know why the game overlay is broken. as a comment states in the mod they cant get a clean texture so they use SetOverlayTexelAspect as a workaround to fix the broken texture, but we currently dont implement that function.
https://github.com/Gistix/portal2vr/blob/4abc0217fbbebae7f978c9f6758ef5dad118dadb/L4D2VR/vr.cpp#L263
we basically have to multiply the width by the TexelAspect

I will fix it tomorrow.

PR #244, opened 2025-12-05
implement XR_MNDX_xdev_space extension

### LINKED ISSUE #27: Introduce full body tracking using the XR_MNDX_xdev_space extension

RinLovesYou:
closes #27 

Tested on VRChat and Resonite

I removed profile map since it was seemingly only used in one place, and devices can get the profile themselves.

### DIFF
--- a/fakexr/src/lib.rs
+++ b/fakexr/src/lib.rs
@@ -1,4 +1,8 @@
 pub mod vulkan;
+
+mod monado_xdev;
+pub use monado_xdev::add_trackers;
+
 use crossbeam_utils::atomic::AtomicCell;
 use glam::{Affine3A, Quat, Vec3};
 use openxr_sys as xr;
@@ -186,6 +190,21 @@ pub unsafe extern "system" fn get_instance_proc_addr(
         ([$($func:tt),+] $pat:pat => $expr:expr) => {
             get_fn!(@arm [$($func),+] -> [] {$pat => $expr})
         };
+        (@arm [ {mndx::$name:ident} $(,$rest:tt)* ] -> [$($arms:tt),*] {$pat:pat => $expr:expr}) => {
+            get_fn!(
+                @arm
+                [$($rest),*] ->
+                [
+                    $($arms,)*
+                    [
+                        x if x == const {
+                            CStr::from_bytes_with_nul_unchecked(concat!("xr", stringify!($name), "\0").as_bytes())
+                        } => Some(std::mem::transmute( paste! { monado_xdev::[<$name:snake>] as openxr_mndx_xdev_space::bindings::$name }))
+                    ]
+                ]
+                {$pat => $expr}
+            )
+        };
         (@arm [$name:ident $(,$rest:tt)*] -> [$($arms:tt),*] {$pat:pat => $expr:expr}) => {
             get_fn!(
                 @arm
@@ -196,21 +215,6 @@ pub unsafe extern "system" fn get_instance_proc_addr(
                         x if x == const {
                             CStr::from_bytes_with_nul_unchecked(concat!("xr", stringify!($name), "\0").as_bytes())
                         } => Some(std::mem::transmute( paste! { [<$name:snake>] as xr::pfn::$name }))
-                    ]
-                ]
-                {$pat => $expr}
-            )
-        };
-        (@arm [ ( mndx :: $name:ident ) $(,$rest:tt)* ] -> [$($arms:tt),*] {$pat:pat => $expr:expr}) => {
-            get_fn!(
-                @arm
-                [$($rest),*] ->
-                [
-                    $($arms,)*
-                    [
-                        x if x == const {
-                            CStr::from_bytes_with_nul_unchecked(concat!("xr", stringify!($name), "\0").as_bytes())
-                        } => Some(std::mem::transmute( paste! { [<$name:snake>] as openxr_mndx_xdev_space::bindings::$name }))
                     ]
                 ]
                 {$pat => $expr}
@@ -252,79 +256,6 @@ pub unsafe extern "system" fn get_instance_proc_addr(
         use vulkan::xr::*;
 
         unsafe {
-            #[cfg(not(feature = "monado"))]
-            {
-                *function = get_fn![[
-                    GetInstanceProcAddr,
-                    CreateInstance,
-                    DestroyInstance,
-                    (EnumerateInstanceExtensionProperties),
-                    (EnumerateApiLayerProperties),
-                    GetVulkanInstanceExtensionsKHR,
-                    GetVulkanDeviceExtensionsKHR,
-                    GetVulkanGraphicsDeviceKHR,
-                    GetVulkanGraphicsRequirementsKHR,
-                    GetSystem,
-                    CreateSession,
-                    DestroySession,
-                    BeginSession,
-                    EndSession,
-                    CreateReferenceSpace,
-                    PollEvent,
-                    DestroySpace,
-                    LocateViews,
-                    RequestExitSession,
-                    (ResultToString),
-                    (StructureTypeToString),
-                    (GetInstanceProperties),
-                    (GetSystemProperties),
-                    CreateSwapchain,
-                    DestroySwapchain,
-                    EnumerateSwapchainImages,
-                    AcquireSwapchainImage,
-                    WaitSwapchainImage,
-                    ReleaseSwapchainImage,
-                    EnumerateSwapchainFormats,
-                    (EnumerateReferenceSpaces),
-                    CreateActionSpace,
-                    LocateSpace,
-                    (EnumerateViewConfigurations),
-                    (EnumerateEnvironmentBlendModes),
-                    (GetViewConfigurationProperties),
-                    (EnumerateViewConfigurationViews),
-                    BeginFrame,
-                    EndFrame,
-                    WaitFrame,
-                    ApplyHapticFeedback,
-                    (StopHapticFeedback),
-                    (PollEvent),
-                    StringToPath,
-                    PathToString,
-                    (GetReferenceSpaceBoundsRect),
-                    GetActionStateBoolean,
-                    GetActionStateFloat,
-                    GetActionStateVector2f,
-                    (GetActionStatePose),
-                    CreateActionSet,
-                    DestroyActionSet,
-                    CreateAction,
-                    DestroyAction,
-                    SuggestInteractionProfileBindings,
-                    AttachSessionActionSets,
-                    GetCurrentInteractionProfile,
-                    SyncActions,
-                    (EnumerateBoundSourcesForAction),
-                    (GetInputSourceLocalizedName)
-                    ]
-
-                    other => {
-                        println!("unknown func: {other:?}");
-                        return xr::Result::ERROR_FUNCTION_UNSUPPORTED;
-                    }
-                ]
-            }
-
-            #[cfg(feature = "monado")]
             {
                 *function = get_fn![[
                     GetInstanceProcAddr,
@@ -387,12 +318,12 @@ pub unsafe extern "system" fn get_instance_proc_addr(
                     SyncActions,
                     (EnumerateBoundSourcesForAction),
                     (GetInputSourceLocalizedName),
-                    (mndx::CreateXDevListMNDX),
-                    (mndx::GetXDevListGenerationNumberMNDX),
-                    (mndx::EnumerateXDevsMNDX),
-                    (mndx::GetXDevPropertiesMNDX),
-                    (mndx::DestroyXDevListMNDX),
-                    (mndx::CreateXDevSpaceMNDX)
+                    {mndx::CreateXDevListMNDX},
+                    {mndx::GetXDevListGenerationNumberMNDX},
+                    {mndx::EnumerateXDevsMNDX},
+                    {mndx::GetXDevPropertiesMNDX},
+                    {mndx::DestroyXDevListMNDX},
+                    {mndx::CreateXDevSpaceMNDX}
                     ]
 
                     other => {
@@ -414,8 +345,8 @@ extern "system" fn enumerate_instance_extension_properties(
     properties: *mut xr::ExtensionProperties,
 ) -> xr::Result {
     assert!(layer_name.is_null());
-    unsafe { *property_count_output = if cfg!(feature = "monado") { 2 } else { 1 } };
-    if property_capacity_input > 1 {
+    unsafe { *property_count_output = 2 };
+    if property_capacity_input >= 2 {
         let props =
             unsafe { std::slice::from_raw_parts_mut(properties, property_capacity_input as usize) };
         props[0] = xr::ExtensionProperties {
@@ -436,13 +367,10 @@ extern "system" fn enumerate_instance_extension_properties(
             extension_version: 1,
         };
 
-        #[cfg(feature = "monado")]
-        {
-            let name = openxr_mndx_xdev_space::XR_MNDX_XDEV_SPACE_EXTENSION_NAME;
-            let name =
-                unsafe { std::slice::from_raw_parts(name.as_ptr() as *const c_char, name.len()) };
-            props[1].extension_name[..name.len()].copy_from_slice(name);
-        }
+        let name = openxr_mndx_xdev_space::XR_MNDX_XDEV_SPACE_EXTENSION_NAME;
+        let name =
+            unsafe { std::slice::from_raw_parts(name.as_ptr() as *const c_char, name.len()) };
+        props[1].extension_name[..name.len()].copy_from_slice(name);
     }
     xr::Result::SUCCESS
 }
@@ -461,7 +389,7 @@ trait XrType {
 
 macro_rules! get_handle {
     ($handle:expr) => {{
-        match <_ as XrType>::to_handle($handle) {
+        match <_ as crate::XrType>::to_handle($handle) {
             Some(handle) => handle,
             None => {
                 eprintln!("unknown handle for {} ({:?})", stringify!($handle), $handle);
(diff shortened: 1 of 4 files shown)


src/input.rs
```diff
@@ -1316,12 +1305,14 @@ impl<C: openxr_data::Compositor> Input<C> {
                 panic!("Failed to create new controller: {:?}", e);
             });
         }
+
+        devices.create_generic_trackers(&self.openxr).unwrap();
```
ImSapphire: This would deadlock if we get an `InteractionProfileChanged` event while polling in `end_session` since we're holding a write lock.
```suggestion
        devices.create_generic_trackers(&self.openxr, session_data).unwrap();
```

src/input/devices.rs
```diff
@@ -207,6 +349,51 @@ impl TrackedDeviceList {
             .map(|(i, _)| i as vr::TrackedDeviceIndex_t)
     }
 
+    pub(super) fn create_generic_trackers(
+        &mut self,
+        xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
+    ) -> xr::Result<()> {
+        let instance = &xr_data.instance;
+        let session_data = xr_data.session_data.get();
+
```
ImSapphire: See above

```suggestion
    pub(super) fn create_generic_trackers(
        &mut self,
        xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
        session_data: &SessionData,
    ) -> xr::Result<()> {
        let instance = &xr_data.instance;

```

src/input.rs
RinLovesYou: Oh! good catch

src/input/devices.rs
```diff
@@ -13,14 +16,21 @@ use super::{profiles::MainAxisType, Input, InteractionProfile};
...
 pub struct TrackedDevice {
     device_type: TrackedDeviceType,
     pub interaction_profile: Option<&'static dyn InteractionProfile>,
     pub profile_path: xr::Path,
     pub connected: bool,
     pub previous_connected: bool,
     pose_cache: Mutex<Option<vr::TrackedDevicePose_t>>,
+
+    // Things generic trackers need to hold on to that can't be Copied
+    xdev: Option<XDev>,
```
ImSapphire: This field is unused.

RinLovesYou:
> `GetLastPoseForTrackedDeviceIndex` should be updated to handle generic trackers, currently it checks if the device_index is HMD or a controller.

You're right, i missed that one

src/input/profiles/vive_tracker.rs
```diff
@@ -0,0 +1,53 @@
...
+    SkeletalInputBindings, StringToPath,
+};
+use crate::{input::legacy::LegacyBindings, openxr_data::Hand};
+use glam::Mat4;
+
+pub struct ViveTracker;
+
+impl InteractionProfile for ViveTracker {
+    fn profile_path(&self) -> &'static str {
+        "/interaction_profiles/valve/index_controller"
```
Supreeeme: Is this intended?

src/input/devices.rs
```diff
@@ -13,14 +16,20 @@ use super::{profiles::MainAxisType, Input, InteractionProfile};
...
     device_type: TrackedDeviceType,
     pub interaction_profile: Option<&'static dyn InteractionProfile>,
     pub profile_path: xr::Path,
     pub connected: bool,
     pub previous_connected: bool,
     pose_cache: Mutex<Option<vr::TrackedDevicePose_t>>,
+
+    // Things generic trackers need to hold on to that can't be Copied
+    xdev_space: Option<xr::Space>,
+    xdev_serial: Option<CString>,
```
Supreeeme: I'd rather the `Copy` and `Clone` derives on `TrackedDeviceType` be removed and these go inside the `GenericTracker` variant. I see no reason why it really needs to be Copy + Clone.

src/input/devices.rs
```diff
@@ -154,6 +288,12 @@ pub struct TrackedDeviceList {
     devices: Vec<TrackedDevice>,
 }
 
+impl Default for TrackedDeviceList {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
Supreeeme: Either `default` or `new` should exist but not both.

src/input/devices.rs
```diff
@@ -207,6 +347,52 @@ impl TrackedDeviceList {
             .map(|(i, _)| i as vr::TrackedDeviceIndex_t)
     }
 
+    pub(super) fn create_generic_trackers(
```
```suggestion
    pub(super) fn create_monado_xdev_trackers(
```

src/input/devices.rs
```diff
@@ -207,6 +347,52 @@ impl TrackedDeviceList {
...
+                TrackedDevice::new(TrackedDeviceType::GenericTracker, None, Some(&ViveTracker));
+
+            tracker.xdev_serial = Some(CString::new(xdev.serial()).unwrap());
+            tracker.xdev_space = Some(xdev.create_space(xr::Posef::IDENTITY).unwrap());
+            tracker.connected = true;
+
+            self.push_device(tracker).unwrap_or_else(|e| {
+                panic!("Failed to create new generic tracker: {:?}", e);
+            });
+        });
```
```suggestion
        let trackers = xdevs.into_iter().map(|xdev| {
            let mut tracker =
                TrackedDevice::new(TrackedDeviceType::GenericTracker, None, Some(&ViveTracker));

            tracker.xdev_serial = Some(CString::new(xdev.serial()).unwrap());
            tracker.xdev_space = Some(xdev.create_space(xr::Posef::IDENTITY).unwrap());
            tracker.connected = true;
            tracker            
        });
        self.devices.extend(trackers);
```

src/input.rs
```diff
@@ -253,6 +236,7 @@ pub struct InputSessionData {
     actions: OnceLock<LoadedActions>,
     estimated_skeleton_actions: OnceLock<SkeletalInputActionData>,
     pose_data: OnceLock<PoseData>,
+    devices: RwLock<TrackedDeviceList>,
```
Supreeeme, line 239: Devices moving to live on the session could use an explanation in the commit message.

src/input.rs
```diff
@@ -597,15 +581,19 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
...
+        let controller_type = self.get_device_string_tracked_property(
+            index,
             vr::ETrackedDeviceProperty::ControllerType_String,
         );
 
         unsafe {
             // Make sure knuckles are always Partial
             // TODO: Remove in favor of using XR_EXT_hand_tracking_data_source
-            if controller_type == Some(c"knuckles") {
+            if controller_type == Some(CString::new("knuckles").unwrap()) {
```
```suggestion
            if controller_type.as_deref() == Some(c"knuckles") {
```

src/system.rs
```diff
@@ -546,20 +546,7 @@ impl vr::IVRSystem023_Interface for System {
...
-                vr::ETrackedDeviceProperty::SerialNumber_String
-                | vr::ETrackedDeviceProperty::ManufacturerName_String
-                | vr::ETrackedDeviceProperty::ControllerType_String => Some(c"<unknown>"),
-                _ => None,
-            },
-            x => input
-                .device_index_to_hand(x)
-                .and_then(|hand| input.get_controller_string_tracked_property(hand, prop)),
-        };
+        let data = input.get_device_string_tracked_property(device_index, prop);
```
Supreeeme: Why are the HMD properties moving into `Input`? The point of having controller (and tracker) properties gated behind `Input` is that the application has to request the Input interface to actually do anything useful with controllers. But this is not the case for HMDs. This would result in games that don't request the input properties not getting any properties at all and probably breaking.

Cargo.toml
```diff
@@ -21,10 +21,14 @@ members = ["openvr", "shaders", "xbuild"]
...
+[patch.crates-io]
+openxr = { git = "https://github.com/ralith/openxrs", rev = "d0afdd3" }
+openxr-sys = { git = "https://github.com/ralith/openxrs", rev = "d0afdd3" }
+
 [workspace.dependencies]
 ash = "0.38.0"
-openxr = { git = "https://github.com/ralith/openxrs", rev = "d0afdd3", version = "0.19.0", features = ["static"] }
-openxr-sys = { git = "https://github.com/ralith/openxrs", rev = "d0afdd3", version = "0.11.0" }
+openxr = { version = "0.19.0", features = ["static"] }
+openxr-sys = { version = "0.11.0" }
```
Supreeeme, line 33: Why?

RinLovesYou: The [xdev crate](https://github.com/Schmarni-Dev/openxr_mndx_xdev_space) uses the openxrs version from crates.io which conflicts when compiling/linking. Patching resolves this conflict.

src/input/devices.rs
RinLovesYou: well, there is one case, like `Input::device_index_to_device_type` where it needs to be copied/cloned. Can't just return it by reference either:

```rust
pub fn device_index_to_device_type(
        &self,
        index: vr::TrackedDeviceIndex_t,
    ) -> Option<&TrackedDeviceType> {
        let session_data = self.openxr.session_data.get();
        let devices = session_data.input_data.devices.read().unwrap();
        let device = devices.get_device(index)?;

        Some(device.get_type()) // cannot return value referencing local variable `devices`
    }
```

RinLovesYou: I've opted for replacing this method with one that directly returns a tracked device class since it was only used in one place

src/system.rs
```diff
@@ -553,12 +553,12 @@ impl vr::IVRSystem023_Interface for System {
...
-                | vr::ETrackedDeviceProperty::ControllerType_String => Some(c"<unknown>"),
+                | vr::ETrackedDeviceProperty::ControllerType_String => {
+                    Some(CString::new("<unknown>").unwrap())
+                }
                 _ => None,
             },
-            x => input
-                .device_index_to_hand(x)
-                .and_then(|hand| input.get_controller_string_tracked_property(hand, prop)),
+            _ => input.get_device_string_tracked_property(device_index, prop),
```
ImSapphire: The above `let Some(input) = self.input.get() else ...` should be removed, then:
```suggestion
            _ => self.input.get().and_then(|input| input.get_device_string_tracked_property(device_index, prop)),
```

src/input/devices.rs
```diff
@@ -70,17 +108,38 @@ fn get_controller_pose(
...
     pub(super) fn new(
         device_type: TrackedDeviceType,
         profile_path: Option<xr::Path>,
         interaction_profile: Option<&'static dyn InteractionProfile>,
     ) -> Self {
         Self {
-            device_type,
             interaction_profile,
             profile_path: profile_path.unwrap_or(xr::Path::NULL),
             connected: device_type == TrackedDeviceType::Hmd,
```
ImSapphire:
```suggestion
            connected: matches!(device_type, TrackedDeviceType::Hmd),
```
Then the `PartialEq` impl for `TrackedDeviceType` can be removed.

src/input/devices.rs
```diff
@@ -120,16 +182,99 @@ impl TrackedDevice {
...
         }
     }
+
+    fn get_string_property(&self, property: vr::ETrackedDeviceProperty) -> Option<&CStr> {
+        let hand = match self.device_type {
+            TrackedDeviceType::Controller { hand } => hand,
+            _ => Hand::Left,
+        };
+
+        let data = self.interaction_profile.as_ref()?.properties();
```
ImSapphire: Is `.as_ref()` required?
```suggestion
        let data = self.interaction_profile?.properties();
```

src/input/devices.rs
```diff
@@ -120,16 +182,99 @@ impl TrackedDevice {
...
+            vr::ETrackedDeviceProperty::ManufacturerName_String => Some(data.manufacturer_name),
+            _ => None,
+        }
+    }
+
+    fn get_int_property(&self, property: vr::ETrackedDeviceProperty) -> Option<i32> {
+        match self.device_type {
+            TrackedDeviceType::Controller { .. } => {
+                let profile = self.interaction_profile?;
+                let data = profile.properties();
```
ImSapphire:
```suggestion
                let data = self.interaction_profile?.properties();
```

src/input/devices.rs
```diff
@@ -120,16 +182,99 @@ impl TrackedDevice {
...
+            }
+            _ => None,
+        }
+    }
+
+    fn get_uint_property(&self, property: vr::ETrackedDeviceProperty) -> Option<u64> {
+        match self.device_type {
+            TrackedDeviceType::Controller { .. } => {
+                let profile = self.interaction_profile?;
+                let data = profile.properties();
```
ImSapphire:
```suggestion
                let data = self.interaction_profile?.properties();
```

src/input/devices.rs
```diff
@@ -1,19 +1,46 @@
...
+        hand: Hand,
+    },
+    #[cfg(feature = "monado")]
+    GenericTracker {
+        space: xr::Space,
+        serial: CString,
+    },
+}
+
+impl Display for TrackedDeviceType {
```
```suggestion
impl std::fmt::Debug for TrackedDeviceType {
```

src/input/devices.rs
```diff
@@ -70,17 +97,39 @@ fn get_controller_pose(
...
+fn get_generic_tracker_pose(
+    xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
+    session_data: &SessionData,
+    tracker: &TrackedDevice,
+    origin: vr::ETrackingUniverseOrigin,
+) -> Option<vr::TrackedDevicePose_t> {
+    let space = match tracker.get_type() {
+        TrackedDeviceType::GenericTracker { space, .. } => Some(space),
+        _ => return None,
+    };
```
```suggestion
    let TrackedDeviceType::GenericTracker { space, .. } = tracker.get_type() else {
        return None,
    };
```

src/input/devices.rs
```diff
@@ -120,16 +173,98 @@ impl TrackedDevice {
...
+            }
+            vr::ETrackedDeviceProperty::TrackingSystemName_String => {
+                Some(data.tracking_system_name)
+            }
+            // Required for controllers to be acknowledged in I Expect You To Die 3
+            vr::ETrackedDeviceProperty::SerialNumber_String => match self.get_type() {
+                TrackedDeviceType::Controller { .. } => Some(*data.serial_number.get(hand)),
+                #[cfg(feature = "monado")]
+                TrackedDeviceType::GenericTracker { serial, .. } => Some(serial.as_c_str()),
+                _ => None,
```
```suggestion
                TrackedDeviceType::Hmd => unreachable!()
```
Force handling if we add any other device types in the future.

src/input/profiles/vive_tracker.rs
```diff
@@ -0,0 +1,53 @@
+use super::{
+    InteractionProfile, MainAxisType, PathTranslation, ProfileProperties, Property,
+    SkeletalInputBindings, StringToPath,
+};
+use crate::{input::legacy::LegacyBindings, openxr_data::Hand};
+use glam::Mat4;
+
+pub struct ViveTracker;
```
Supreeeme, line 8: A basic test testing the ability to get i.e. positions and some properties (doesn't have to be all of them) from a tracker would be nice.

src/input/devices.rs
```diff
@@ -120,16 +171,98 @@ impl TrackedDevice {
...
+            }
+            vr::ETrackedDeviceProperty::TrackingSystemName_String => {
+                Some(data.tracking_system_name)
+            }
+            // Required for controllers to be acknowledged in I Expect You To Die 3
+            vr::ETrackedDeviceProperty::SerialNumber_String => match self.get_type() {
+                TrackedDeviceType::Controller { .. } => Some(*data.serial_number.get(hand)),
+                #[cfg(feature = "monado")]
+                TrackedDeviceType::GenericTracker { serial, .. } => Some(serial.as_c_str()),
+                _ => unreachable!(),
```
```suggestion
                TrackedDeviceType::Hmd => unreachable!(),
```
The point is that if there are other device types added this is a compiler error to not handle it.

src/input/profiles/vive_tracker.rs
RinLovesYou: Yes i thought so too but I'm not sure how that would look in fakexr. Do i implement the extension there?

Supreeeme: Yeah, you'll have to add it to `enumerate_instance_extension_properties` and add the functions it needs to `get_instance_proc_addr`.

GK-Gaming:
This works great! But i cant see the hand/controller 3d model when calibrating, so its kinda difficult to do...

SpookySkeletons:
> This works great! But i cant see the hand/controller 3d model when calibrating, so its kinda difficult to do...

This is a separate issue which requires the implementation of controller 3D models fed to the application over the proper OpenVR interface. OpenComposite still uses placeholder hands for instance. You could open an issue for this if you like but it's out of scope of this PR.

src/input/profiles/vive_tracker.rs
Supreeeme: I plan to do a new release soon, and I'd like to have this PR in the new release. Do you think you'll have tests ready soon? If not, I can write them.

RinLovesYou: Yep, sorry i couldn't last week.

RinLovesYou:
> > This works great! But i cant see the hand/controller 3d model when calibrating, so its kinda difficult to do...
> 
> This is a separate issue which requires the implementation of controller 3D models fed to the application over the proper OpenVR interface. OpenComposite still uses placeholder hands for instance. You could open an issue for this if you like but it's out of scope of this PR.

Yeah there is a repo of models that we should be able to use but that'll be another PR

fakexr/src/lib.rs
```diff
@@ -201,6 +201,21 @@ pub unsafe extern "system" fn get_instance_proc_addr(
...
+        (@arm [ ( mndx :: $name:ident ) $(,$rest:tt)* ] -> [$($arms:tt),*] {$pat:pat => $expr:expr}) => {
+            get_fn!(
+                @arm
+                [$($rest),*] ->
+                [
+                    $($arms,)*
+                    [
+                        x if x == const {
+                            CStr::from_bytes_with_nul_unchecked(concat!("xr", stringify!($name), "\0").as_bytes())
+                        } => Some(std::mem::transmute( paste! { [<$name:snake>] as openxr_mndx_xdev_space::bindings::$name }))
```
Supreeeme: Instead of duplicating the entire function list I think you should be able to just add the `#[cfg(feature = monado)]` on the arm here. Also, maybe I should make this macro less cursed.

fakexr/src/lib.rs
```diff
@@ -1770,3 +1892,165 @@ extern "system" fn apply_haptic_feedback(
 
     xr::Result::SUCCESS
 }
+
+#[cfg(feature = "monado")]
+#[derive(Default)]
+struct XDevListMNDX {
+    generation_number: u64,
+    xdevs: Vec<XDev>,
+}
```
Supreeeme: Maybe all the monado related structs and functions should go in a separate submodule.

src/input/devices.rs
```diff
@@ -258,119 +452,147 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+        };
+
+        // we need to wait two frames for the tracker to be connected.
+        frame();
+        assert!(f.input.device_index_to_tracked_device_class(2).is_none());
+        frame();
+        assert!(
+            f.input.device_index_to_tracked_device_class(2)
+                == Some(vr::ETrackedDeviceClass::GenericTracker)
+        );
```
Supreeeme, line 566: It seems like trackers are automatically created in tests? I'd prefer if there was an explicit function that fakexr exposes that enables trackers.

src/input/devices.rs
```diff
@@ -258,119 +452,147 @@ impl<C: openxr_data::Compositor> Input<C> {
...
-        })
+        let session_data = self.openxr.session_data.get();
+        let devices = session_data.input_data.devices.read().unwrap();
+        let device = devices.get_device(index)?;
+
+        device.get_uint_property(property)
+    }
+}
+
+#[cfg(test)]
```
Supreeeme, line 541:
```suggestion
#[cfg(all(test, monado))]
```

RinLovesYou: i think it would be `#[cfg(all(test, feature = "monado"))]`

src/input/devices.rs
RinLovesYou: right, I think `Input` would need to expose `create_monado_generic_trackers()` for this though and at that point we might as well just directly call that from the tests instead of going through fakexr. What do you think?

Supreeeme: Why would you need to expose that? I'm just saying to have `enumerate_x_devs_m_n_d_x` return 0 devices until some fakexr function (i.e. `fakexr::add_trackers`) is called.

RinLovesYou: Oh i see

Supreeeme:
In the interest of moving this forward I went ahead and made some of the test adjustments myself.

PR #245, opened 2025-12-05
Implement poses functions with "seconds from now"

xytovl:
OpenXR doesn't natively have a concept of "now", but has extensions to convert OS time to XrTime so that we can implement relative time ourselves.

GetPoseActionDataRelativeToNow is used for instance in The Lab, the change does however not fix the hand movement which is still not smooth.

### DIFF
--- a/src/input/devices.rs
+++ b/src/input/devices.rs
@@ -215,6 +215,7 @@ impl<C: openxr_data::Compositor> Input<C> {
         &self,
         poses: &mut [vr::TrackedDevicePose_t],
         origin: Option<vr::ETrackingUniverseOrigin>,
+        xr_time: xr::Time,
     ) {
         tracy_span!();
         let devices = self.devices.read().unwrap();
@@ -226,7 +227,7 @@ impl<C: openxr_data::Compositor> Input<C> {
             if let Some(device) = device {
                 *pose = device
                     .get_pose(
-                        self.openxr.display_time.get(),
+                        xr_time,
                         &session_data,
                         origin.unwrap_or(session_data.current_origin),
                     )
--- a/src/system.rs
+++ b/src/system.rs
@@ -820,7 +820,7 @@ impl vr::IVRSystem023_Interface for System {
     fn GetDeviceToAbsoluteTrackingPose(
         &self,
         origin: vr::ETrackingUniverseOrigin,
-        _seconds_to_photon_from_now: f32,
+        seconds_to_photon_from_now: f32,
         pose_array: *mut vr::TrackedDevicePose_t,
         pose_count: u32,
     ) {
@@ -829,6 +829,7 @@ impl vr::IVRSystem023_Interface for System {
             .get_poses(
                 unsafe { std::slice::from_raw_parts_mut(pose_array, pose_count as usize) },
                 Some(origin),
+                self.openxr.time_from_now(seconds_to_photon_from_now),
             );
     }
     fn SetDisplayVisibility(&self, _: bool) -> bool {
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -861,7 +861,7 @@ impl vr::IVRCompositor029_Interface for Compositor {
         };
         self.input
             .force(|_| Input::new(self.openxr.clone()))
-            .get_poses(render_poses, None);
+            .get_poses(render_poses, None, self.openxr.display_time.get());
 
         // Not entirely sure how the game poses are supposed to differ from the render poses,
         // but a lot of games use the game pose array for controller positions.


xytovl:
I could test GetPoseActionDataRelativeToNow in The Lab, which doesn't make any visible change.
I don't know of games that use GetDeviceToAbsoluteTrackingPose, so I'm not sure how to test that change.

src/input.rs
```diff
@@ -721,153 +722,38 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
...
-        _seconds_from_now: f32,
+        seconds_from_now: f32,
         action_data: *mut vr::InputPoseActionData_t,
         action_data_size: u32,
         restrict_to_device: vr::VRInputValueHandle_t,
     ) -> vr::EVRInputError {
-        self.GetPoseActionDataForNextFrame(
+        let xr_time = match self.openxr.instance.now() {
+            Ok(time) => {
+                xr::Time::from_nanos(time.as_nanos() + (seconds_from_now * 1000000000.) as i64)
```
```suggestion
                xr::Time::from_nanos(time.as_nanos() + std::time::Duration::from_secs_f32(seconds_from_now).as_nanos() as i64)
```

src/input/devices.rs
```diff
@@ -82,33 +77,36 @@ impl TrackedDevice {
...
     }
 
     pub fn get_pose(
         &self,
-        xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
+        xr_time: xr::Time,
         session_data: &SessionData,
         origin: vr::ETrackingUniverseOrigin,
     ) -> Option<vr::TrackedDevicePose_t> {
+        let nanos = xr_time.as_nanos();
```
Supreeeme, line 90: Why not have the pose cache be `Mutex<HashMap<xr::Time, TrackedDevicePose>>`?

src/input/devices.rs
```diff
@@ -82,33 +77,36 @@ impl TrackedDevice {
...
             TrackedDeviceType::Controller { .. } => {
-                get_controller_pose(xr_data, session_data, self, origin)
+                get_controller_pose(xr_time, session_data, self, origin)
             }
+        } {
+            pose_cache.insert(nanos, pose);
         };
 
-        *pose_cache
+        pose_cache.get(&nanos).copied()
```
```suggestion
        let pose = match self.device_type {
            TrackedDeviceType::Hmd => get_hmd_pose(xr_time, session_data, origin),
            TrackedDeviceType::Controller { .. } => {
                get_controller_pose(xr_time, session_data, self, origin)
            }
        };

        pose.and_then(|p| pose_cache.entry(xr_time).insert_entry(p).get())
```

src/system.rs
```diff
@@ -816,15 +820,22 @@ impl vr::IVRSystem023_Interface for System {
...
+        seconds_to_photon_from_now: f32,
         pose_array: *mut vr::TrackedDevicePose_t,
         pose_count: u32,
     ) {
+        let xr_time = match self.openxr.instance.now() {
+            Ok(time) => xr::Time::from_nanos(
+                time.as_nanos() + (seconds_to_photon_from_now * 1000000000.) as i64,
+            ),
+            _ => self.openxr.display_time.get(),
+        };
```
Supreeeme: Since this is used in multiple places I think it's worth throwing it into a helper function on `OpenXrData`.

src/input/devices.rs
xytovl: It complained that xr::Time was not hashable:

```
   Compiling xrizer v0.3.0 (/home/xytovl/projects/xrizer)
error[E0599]: the method `get` exists for struct `std::sync::MutexGuard<'_, HashMap<openxr::Time, TrackedDevicePose_t>>`, but its trait bounds were not satisfied
  --> src/input/devices.rs:91:40
   |
91 |         if let Some(pose) = pose_cache.get(xr_time) {
   |                                        ^^^ method cannot be called due to unsatisfied trait bounds
   |
  ::: /home/xytovl/.cargo/git/checkouts/openxrs-419b52284129469d/d0afdd3/sys/src/lib.rs:46:1
   |
46 | pub struct Time(i64);
   | --------------- doesn't satisfy `openxr::Time: Hash`
   |
   = note: the following trait bounds were not satisfied:
           `openxr::Time: Hash`
```

src/input/devices.rs
xytovl: Done, with 2 small changes: the xr_time is still i64 due to the hash thing. I also had to add a `Some()` to wrap the result.

xytovl:
> This could use a test if you're up to writing one, if not I'll write it myself.

What would you test? The cache mechanism, relative time, fallback if openxr now is not available?
I am not familiar with rust yet, so I'm not sure how to mock the instance method.

src/input/devices.rs
Supreeeme: Ah. In this case I'd prefer if a type like this was defined:
```rust
#[derive(PartialEq, Eq, derive_more::From, Copy, Clone)]
struct Time(xr::Time);

impl Hash for Time {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.0.as_nanos().hash(state);
    }
}
```

Then you should be able to use it like:
```rust
pose_cache.get(&Time(xr_time))
pose_cache.insert(Time(xr_time))
// etc
```

Supreeeme:
> What would you test?

Relative time. Check that the poses returned from GetPoseActionDataRelativeToNow( and GetPoseActionDataForNextFrame actually differ, and that it returns the right pose basically.

The way xrizer handles tests is primarily unit testing each interface with a mocked OpenXR runtime, this fake runtime lives in the `fakexr` directory. If you look at the code in there you can see it's just implementing all of the relevant OpenXR functions from the runtime side. So you can add whatever helpers you need and just call the OpenVR functions like an application would. The input tests specifically live in `src/input/tests.rs`, they should be insightful at least.

Supreeeme:
Sorry I keep accidentally closing it lol

PR #257, opened 2025-12-23
Allow any device to be used as a generic tracker

coolGi69:
Also see https://github.com/ImSapphire/xrizer/pull/1.

This is a simple PR that adds the `XRIZER_TRACKER_SERIALS` environment variable to select a semicolon separated list of devices that the user wants as generic trackers (similar to `OPENCOMPOSITE_TRACKER_SERIALS`).\
It can be used to map controllers as extra tracker points.

Tested with VRChat & Resonite.\
If there's anything wrong with the code (i.e., naming) or anything, I'm happy to fix it!


coolGi69:
Sorry, accidentally pressed the close button. 😅

PR #260, opened 2025-12-26
Stub out a pair of overlay transform functions (for PAYDAY 2 VR)

demif1end:
This is necessary for PAYDAY 2 VR. The game works flawlessly after this patch is made, though it should be noted that, probably due to the nature of these functions, wlx-overlay-s acted a little weird (as if it's transform was... offset... in some manner. if anyone can check if it was just my overlay acting up, i would appreciate it!)
Besides that, the game works, but these functions will probably need implemented eventually...


galister:
WlxOverlay-S doesn't use xrizer, so no way that it's related.

PR #261, opened 2025-12-28
input: fix hand tracking with no interaction profile and/or grip pose

### LINKED ISSUE #165: [Bug]: Hand tracking doesn't work when no grip pose is provided
### Operating System

NixOS Xantusia (25.11)

### OpenXR Runtime

WiVRn

### Describe the issue.

WiVRn since v25.8 doesn't provide aim, grip, or palm poses for hands, only XR_EXT_hand_tracking data. xrizer doesn't seem to handle this case correctly.

### Steps to reproduce

_No response_

ImSapphire:
Fixes hand tracking on WiVRn (WiVRn/WiVRn@1e2823c0196971caa9b32956a2bd0469b3c9da76). Closes #165 

Hand tracking is still pretty scuffed, we don't support the 'augmented' inputs from hand tracking that SteamVR supports (/input/[finger]_pinch trigger input, /input/index_point trigger input, /input/system left side button input, plus rebinds for basic trigger and grip pull).

I don't like the `pose_is_synthesised` hack, but this is what OC did and I'm not good enough at math to figure out how to fix the wrist pose when we're using a fake grip.

Tested in VRChat and Resonite.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -56,7 +56,7 @@ pub struct Input<C: openxr_data::Compositor> {
     set_map: RwLock<SlotMap<ActionSetKey, String>>,
     loaded_actions_path: OnceLock<PathBuf>,
     legacy_state: legacy::LegacyState,
-    skeletal_tracking_level: RwLock<vr::EVRSkeletalTrackingLevel>,
+    skeletal_tracking_level: RwLock<[vr::EVRSkeletalTrackingLevel; 2]>,
     skeletal_bone_cache:
         [Mutex<Option<[vr::VRBoneTransform_t; HandSkeletonBone::Count as usize]>>; 2],
     estimated_finger_state: [Mutex<FingerState>; 2],
@@ -128,7 +128,7 @@ impl<C: openxr_data::Compositor> Input<C> {
             left_hand_key,
             right_hand_key,
             legacy_state: Default::default(),
-            skeletal_tracking_level: RwLock::new(vr::EVRSkeletalTrackingLevel::Estimated),
+            skeletal_tracking_level: RwLock::new([vr::EVRSkeletalTrackingLevel::Estimated; 2]),
             skeletal_bone_cache: Default::default(),
             estimated_finger_state: [
                 Mutex::new(FingerState::new()),
@@ -609,7 +609,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
             if controller_type.as_deref() == Some(c"knuckles") {
                 *level = vr::EVRSkeletalTrackingLevel::Partial;
             } else {
-                *level = *self.skeletal_tracking_level.read().unwrap();
+                *level = self.skeletal_tracking_level.read().unwrap()[*hand as usize - 1];
             }
         }
         vr::EVRInputError::None
--- a/src/input/skeletal.rs
+++ b/src/input/skeletal.rs
@@ -182,8 +182,8 @@ impl<C: openxr_data::Compositor> Input<C> {
         *bone_cache = Some(transforms);
 
         finalize_transforms(&mut transforms, space);
-
-        *self.skeletal_tracking_level.write().unwrap() = vr::EVRSkeletalTrackingLevel::Full;
+        self.skeletal_tracking_level.write().unwrap()[hand as usize - 1] =
+            vr::EVRSkeletalTrackingLevel::Full;
 
         Some(transforms)
     }
@@ -349,7 +349,8 @@ impl<C: openxr_data::Compositor> Input<C> {
         *bone_cache = Some(transforms);
 
         finalize_transforms(&mut transforms, space);
-        *self.skeletal_tracking_level.write().unwrap() = vr::EVRSkeletalTrackingLevel::Estimated;
+        self.skeletal_tracking_level.write().unwrap()[hand as usize - 1] =
+            vr::EVRSkeletalTrackingLevel::Estimated;
 
         transforms
     }


src/input/devices.rs
```diff
@@ -71,26 +84,109 @@ fn get_controller_pose(
...
+                let position = bone.pose.position;
+                let orientation = bone.pose.orientation;
+                let mut pose = Mat4::from_rotation_translation(
+                    Quat::from_xyzw(orientation.x, orientation.y, orientation.z, orientation.w),
+                    Vec3::from_array([position.x, position.y, position.z]),
+                );
+
+                let mut offset = Mat4::from_translation(match data.hand {
+                    Hand::Left => Vec3::from_array([0.09, -0.03, -0.09]),
+                    Hand::Right => Vec3::from_array([-0.09, -0.03, -0.09]),
```
Supreeeme, line 118: Where are these values coming from?

src/input/devices.rs
```diff
@@ -71,26 +84,109 @@ fn get_controller_pose(
...
+                        -FRAC_PI_4
+                    } else {
+                        FRAC_PI_4
+                    },
+                    if data.hand == Hand::Left {
+                        -FRAC_PI_2
+                    } else {
+                        FRAC_PI_2
+                    },
+                );
```
Supreeeme, line 133: What's up with this transform?

src/input/devices.rs
```diff
@@ -71,26 +84,109 @@ fn get_controller_pose(
...
+                }
+            };
+
+            data.pose_is_synthesised.store(true, Ordering::Relaxed);
+            (
+                xr::SpaceLocation {
+                    location_flags: bone.location_flags,
+                    pose,
+                },
+                xr::SpaceVelocity::default(),
```
Supreeeme: This is not great. No velocity on hands results in stuff like throwing things not working. At the very least a `TODO` comment should go here.

src/input/devices.rs
```diff
@@ -175,20 +311,45 @@ impl TrackedDevice {
...
+            if matches!(self.device_type, TrackedDeviceType::Controller { .. })
+                && self
+                    .get_hand_skeleton(xr_data, session_data.tracking_space())
+                    .is_some()
+            {
+                Some(&VRLinkHand)
+            } else {
+                None
+            }
+        })
```
Supreeeme, line 337: Is hand tracking only something that can be dynamically switched to, and does the runtime change the interaction profiles when hand tracking is enabled? If the profile doesn't change this will be stuck with the old profile and never work.

src/input/devices.rs
```diff
@@ -287,10 +462,47 @@ pub struct TrackedDeviceList {
...
+        let create_hand_tracker = |hand: Hand| {
+            session
+                .create_hand_tracker(hand.into())
+                .inspect_err(|e| {
+                    if !matches!(
+                        *e,
+                        xr::sys::Result::ERROR_EXTENSION_NOT_PRESENT
+                            | xr::sys::Result::ERROR_FEATURE_UNSUPPORTED
+                    ) {
+                        log::warn!("Failed to create hand tracker for hand {hand:?}: {e}");
```
```suggestion
                        log::debug!("Failed to create hand tracker for hand {hand:?}: {e}");
```

src/input/devices.rs
```diff
@@ -287,10 +462,47 @@ pub struct TrackedDeviceList {
...
+                TrackedDevice::new(
+                    TrackedDeviceType::Controller(ControllerData {
+                        hand: Hand::Right,
+                        hand_tracker: create_hand_tracker(Hand::Right),
+                        skeleton_cache: Mutex::new(Box::new(None)),
+                        pose_is_synthesised: false.into(),
+                    }),
+                    None,
+                    None,
+                ),
```
Supreeeme, line 512: Why are controllers being automatically created at startup?

src/input.rs
```diff
@@ -676,26 +680,17 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
...
         };
-        let pose_data = data.input_data.pose_data.get().unwrap();
         unsafe {
-            std::ptr::addr_of_mut!((*action_data).bActive).write(
-                pose_data
-                    .grip
-                    .is_active(&data.session, xr::Path::NULL)
-                    .unwrap(),
-            );
+            std::ptr::addr_of_mut!((*action_data).bActive).write(true);
```
Supreeeme: This shouldn't be unconditionally active.

src/input/devices.rs
```diff
@@ -13,13 +18,21 @@ use crate::openxr_data::{self, Hand, OpenXrData, SessionData};
...
+use super::{
+    profiles::{vrlink_hand::VRLinkHand, MainAxisType},
+    Input, InteractionProfile,
+};
+
+pub struct ControllerData {
+    pub hand: Hand,
+    pub hand_tracker: Option<xr::HandTracker>,
+    pub skeleton_cache: Mutex<Box<Option<xr::HandJointLocations>>>,
+    pub pose_is_synthesised: AtomicBool,
```
Supreeeme: You could instead have an enum like:
```
enum SkeletonPose {
    Real(Box<Option<xr::HandJointLocations>>),
    Synthesized(Box<Option<xr::HandJointLocations>>)
}
```
and then just have a field like:
```
skeleton_cache: Mutex<SkeletonPose>
```

Also, I don't think all of the fields in this struct need to be `pub`.

src/input.rs
ImSapphire: i think i did that for testing and forgot to revert, oops

src/input/devices.rs
ImSapphire: without this hand tracking will not work if controllers aren’t connected at least once

src/input/devices.rs
ImSapphire: at least on Quest through WiVRn the interaction profile will change to NULL when switching to hand tracking.

i don’t think it’s critical that we return the VRLinkHand profile in all cases where hand tracking is active, i just did this so hand tracking works with no interaction profile

src/input/devices.rs
ImSapphire: they're pulled from OpenComposite

src/input/devices.rs
ImSapphire: it's from OC. without this transform, when the fingertips are facing forward and palms are facing down, in-game the palms will be facing inward and fingers are pointed 45deg down

src/input/devices.rs
SpookySkeletons: These values derived from OC, are pulled from SteamVR's SteamLink for hands, so they should be spot on.

src/input/devices.rs
SpookySkeletons: Most likely also derived from SteamLink ditto the above comment by me.

src/input/devices.rs
SpookySkeletons: If this came from OC, definitely pass velocity, we have problems throwing in OC games as this might have been pulled in from that and should not be `TODO` if we can help it.

src/input/devices.rs
ImSapphire: synthesised pose doesn't really have anything to do with the openxr hand skeleton though, so i'm not sure it would make sense to wrap this in an enum here

src/input/devices.rs
ImSapphire: added velocity for synthesised pose

src/input/devices.rs
Supreeeme: This could use a comment or be a `const` or something.

src/input/devices.rs
Supreeeme: Can this be a comment?

src/input/devices.rs
```diff
@@ -71,26 +84,116 @@ fn get_controller_pose(
...
     session_data: &SessionData,
     controller: &TrackedDevice,
+    data: &ControllerData,
     origin: vr::ETrackingUniverseOrigin,
 ) -> Option<vr::TrackedDevicePose_t> {
+    // Try to synthesise a grip pose from palm joint
+    let pose_from_hand_tracking = || {
+        if let Some((joints, velocities)) =
+            controller.get_hand_skeleton(xr_data, session_data.tracking_space())
+        {
```
Supreeeme, line 96: This could be a method on `ControllerData`.

src/input/devices.rs
```diff
@@ -158,8 +261,48 @@ impl TrackedDevice {
...
     }
 
+    pub fn get_hand_skeleton(
+        &self,
+        xr_data: &OpenXrData<impl crate::openxr_data::Compositor>,
+        base: &xr::Space,
+    ) -> Option<(xr::HandJointLocations, xr::HandJointVelocities)> {
+        let TrackedDeviceType::Controller(data) = self.get_type() else {
+            return None;
+        };
```
Supreeeme, line 273: This should just be a method on ControllerData.

src/input/devices.rs
Supreeeme: Why? We should not report controllers if they don't actually exist. If I'm understanding the VRLinkHand profile properly, we just need to add bindings for the hand_interaction profile. In fact, couldn't we just do that instead of checking for the hand skeleton?

src/input/skeletal.rs
```diff
@@ -11,52 +12,75 @@ use openxr::{self as xr};
...
-    /// Returns false if hand tracking data couldn't be generated for some reason.
     pub(super) fn get_bones_from_hand_tracking(
         &self,
         session_data: &SessionData,
         space: vr::EVRSkeletalTransformSpace,
-        hand_tracker: &xr::HandTracker,
         hand: Hand,
-        transforms: &mut [vr::VRBoneTransform_t],
-    ) {
+    ) -> Option<[vr::VRBoneTransform_t; Count as usize]> {
```
Supreeeme, line 24: Why is this changed?

src/input/devices.rs
ImSapphire: > If I'm understanding the VRLinkHand profile properly, we just need to add bindings for the hand_interaction profile. In fact, couldn't we just do that instead of checking for the hand skeleton?

It's possible to have XR_EXT_hand_tracking data available while XR_EXT_hand_interaction is not supported, I only added the VRLinkHand profile for the properties

Supreeeme: Fair enough, but that should be a comment on the profile, that's not really immediately obvious. In any case though, we should not create controllers when they don't exist, that's just confusing. There should be a way to check if hand tracking is available and try to pull a pose from it without this. `XrSystemHandTrackingPropertiesEXT::supportsHandTracking` might be useful.

ImSapphire: > In any case though, we should not create controllers when they don't exist, that's just confusing

I do agree, but there's lots of functions that rely on the controllers existing, like IsTrackedDeviceConnected and Get*TrackedDeviceProperty, so I'm pretty sure we have to do this for hand tracking to work without having gotten a valid interaction profile for the hands, right?

Supreeeme: It would make more sense to take an approach similar to the info set used to determine if there are controllers before reporting them to the game - each frame, you could attempt to get a pose from the hand tracker, and when it succeeds you can add a controller. I don't have a real example on this but I'm pretty sure lying about controllers being connected can actually mess some games up. (should've written a test, rip)

PR #268, opened 2026-01-05
don't use Vec::leak

sofoxe1:
:3

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -385,64 +385,88 @@ impl vr::IVRSystem023_Interface for System {
         eye: vr::EVREye,
         ty: vr::EHiddenAreaMeshType,
     ) -> vr::HiddenAreaMesh_t {
-        if !self.openxr.enabled_extensions.khr_visibility_mask {
-            return Default::default();
+        #[derive(Default)]
+        //same as HiddenAreaMesh_t but with rust types
+        struct HiddenAreaMesh {
+            pub vertex_data: Option<Box<[openvr::HmdVector2_t]>>,
+            pub triangle_count: u32,
         }
-
-        debug!("GetHiddenAreaMesh: area mesh type: {ty:?}");
-        let mask_ty = match ty {
-            vr::EHiddenAreaMeshType::Standard => xr::VisibilityMaskTypeKHR::HIDDEN_TRIANGLE_MESH,
-            vr::EHiddenAreaMeshType::Inverse => xr::VisibilityMaskTypeKHR::VISIBLE_TRIANGLE_MESH,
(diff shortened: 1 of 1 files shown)


src/system.rs
```diff
@@ -385,64 +385,90 @@ impl vr::IVRSystem023_Interface for System {
...
+        #[derive(Default)]
+        //same as HiddenAreaMesh_t but with rust types
+        struct HiddenAreaMesh {
+            pub vertex_data: Option<Box<[openvr::HmdVector2_t]>>,
+            pub triangle_count: u32,
         }
+        //steamvr keeps the cache in the class field
+        //
+        //ClearHiddenAreaCache is exported under CVRSystemLatest but since it's not documented, i'm gonna assume it's not used by anything externally
+        static mut HIDDEN_AREA_CACHE: [HiddenAreaMesh; 2] = unsafe { core::mem::zeroed() };
```
Supreeeme, line 397: This should live inside of `System` and doesn't need to be a `static mut`, it can be a `OnceCell`

src/system.rs
```diff
@@ -385,64 +385,90 @@ impl vr::IVRSystem023_Interface for System {
...
+        if unsafe { HIDDEN_AREA_CACHE[0].vertex_data.is_none() } {
+            //steamvr does calculation for both eyes at once
+            for eye in [vr::EVREye::Left, vr::EVREye::Right] {
+                if !self.openxr.enabled_extensions.khr_visibility_mask {
+                    return Default::default();
+                }
```
Supreeeme: This can move outside the loop

src/system.rs
sofoxe1: it would have to be a OnceLock so rust doesn't complain about Sync so using unsafe is theoretically faster as it doesn't use atomic instructions:)

sofoxe1:
> This change is fine, but is this actually fixing something?

nope

src/system.rs
sofoxe1: it's only used by this function and putting in it system would add one more entry to autocomplete

Supreeeme: Putting it in System makes it actually testable, and I think is generally clearer.

sofoxe1:
so i got to rewriting this PR and after reading the decompiled vrclient source code again steamvr actually caches data for all combinations of types and eye (for some reason i assumed application would not request different mesh types)  since it doesn't cause any problems and applications only seem to call this function twice at the start i think current implementation is actually better then exactly copying steamvr

the only thing that could be added is this (check eye and ty value) but that's probably not a real issue either
image

PR #270, opened 2026-01-06
Support IVRInput_004

ImSapphire:
Required by I Expect You To Die for controllers to work.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -27,7 +27,7 @@ use openvr as vr;
 use openxr as xr;
 use slotmap::{new_key_type, Key, KeyData, SecondaryMap, SlotMap};
 use std::collections::{HashMap, HashSet, VecDeque};
-use std::ffi::{c_char, CStr, CString};
+use std::ffi::{c_char, c_void, CStr, CString};
 use std::mem::ManuallyDrop;
 use std::ops::Deref;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


src/input.rs
```diff
@@ -1237,6 +1237,118 @@ impl<C: openxr_data::Compositor> vr::IVRInput005On006 for Input<C> {
...
+        &self,
+        compressed_buffer: *mut std::os::raw::c_void,
+        compressed_buffer_size: u32,
+        transform_space: *mut vr::EVRSkeletalTransformSpace,
+        transform_array: *mut vr::VRBoneTransform_t,
+        transform_array_count: u32,
+    ) -> vr::EVRInputError {
+        let space =
+            unsafe { transform_space.as_ref() }.unwrap_or(&vr::EVRSkeletalTransformSpace::Parent);
+        <Self as vr::IVRInput010_Interface>::DecompressSkeletalBoneData(
```
Supreeeme: This looks kinda funny. Typically the mut pointers are used as outputs. It's possible that was just an API goof, but I'd rather not get random segfaults. I'm not positive what to do here - I'd be inclined to just mark this `todo` since we don't even implement this yet anyway.

src/input.rs
```diff
@@ -1237,6 +1237,118 @@ impl<C: openxr_data::Compositor> vr::IVRInput005On006 for Input<C> {
...
+            transform_array,
+            transform_array_count,
+        )
+    }
+
+    #[inline]
+    fn GetOriginLocalizedName(
+        &self,
+        origin: vr::VRInputValueHandle_t,
+        name_array: *mut std::os::raw::c_char,
```
Supreeeme: Can you change this (and other instances of it) to just `c_char`? It's already imported.

src/input.rs
ImSapphire: I believe it's supposed to be an input since it was changed in later API versions, but I'll just mark it todo.

PR #289, opened 2026-01-27
Add OpenVR v1.0.1 for Rez Infinite

prikhi:
REZ Infinite requires `IVRCompositor_015`, but went ahead & pulled in all unique interface versions from v1.0.1

Tested & game starts up but haven't played any yet.


prikhi:
Played it a bit, not perfect - haptics seems to be permanently going off and trigger is finicky, but _was_ able to play...

Issue #255, opened 2025-12-22
[Bug]: Audio Trip: No Haptics previously, no controllers in latest nightly


alpnist:
### Operating System

Xubuntu 24.04

### OpenXR Runtime

WiVRn

### Describe the issue.

**Please attach a log if applicable.**
Logs are located at `~/.local/state/xrizer/xrizer.txt`.
[2025-12-22T15:18:13.973 INFO  xrizer ThreadId(1)] Initializing XRizer version 0.4.0
[2025-12-22T15:18:13.973 INFO  xrizer::clientcore ThreadId(1)] Creating ClientCore version "IVRClientCore_003"
[2025-12-22T15:18:14.227 INFO  xrizer::openxr_data ThreadId(1)] Creating OpenXR session with graphics API openxr::graphics::vulkan::Vulkan
[2025-12-22T15:18:14.229 INFO  xrizer::openxr_data ThreadId(1)] New session created!
[2025-12-22T15:18:14.229 INFO  xrizer::openxr_data ThreadId(1)] OpenXR session state changed: READY
[2025-12-22T15:18:14.229 INFO  xrizer::openxr_data ThreadId(1)] Began OpenXR session.
[2025-12-22T15:18:14.230 INFO  xrizer::input ThreadId(1)] loading action manifest from "/home/alpinist/.steam/debian-installation/steamapps/common/Audio Trip/AudioTrip_Data/StreamingAssets/SteamVR/actions.json"
[2025-12-22T15:18:14.271 INFO  xrizer::input::action_manifest ThreadId(1)] loading bindings for /interaction_profiles/htc/vive_controller
[2025-12-22T15:18:14.271 INFO  xrizer::input::action_manifest ThreadId(1)] loading bindings for /interaction_profiles/khr/simple_controller
[2025-12-22T15:18:14.271 WARN  xrizer::input::action_manifest ThreadId(1)] Action for invalid path /user/hand/left/input/grip/click for /interaction_profiles/khr/simple_controller, ignoring (/actions/audiotrip/in/grip)
[2025-12-22T15:18:14.271 WARN  xrizer::input::action_manifest ThreadId(1)] Action for invalid path /user/hand/right/input/grip/click for /interaction_profiles/khr/simple_controller, ignoring (/actions/audiotrip/in/grip)
[...]

PR #292, opened 2026-01-31
[#255] Ensure Devices Are Reinitialized on Session Reboot

prikhi:
Modify the Input system's `load_action_manifest` method to call it's `interaction_profile_changed` method, which ensures we re-initialize the `TrackedDeviceList` after forcing a new interaction profile.

This change cascades to the OpenXrData's `restart_session` method, ensuring devices continue to be tracked as soon as possible when a session is restarted.

This fixes an issue with controller positions being stuck at the headset origin in Audio Trip, as after a session start it requests device properties. Without this change, the `xrizer` session does not yet know about the devices and fails the "get property" calls, and the hand positions in game remain stuck.

Refs #255

### DIFF
--- a/src/input/action_manifest.rs
+++ b/src/input/action_manifest.rs
@@ -171,6 +171,7 @@ impl<C: openxr_data::Compositor> Input<C> {
             .session
             .sync_actions(&[xr::ActiveActionSet::new(&info_set)])
             .unwrap();
+        self.interaction_profile_changed(session_data);
 
         // Transform actions and sets into maps
         // If the application has already requested the handle for an action/set, we need to


src/input/action_manifest.rs
```diff
@@ -171,6 +171,7 @@ impl<C: openxr_data::Compositor> Input<C> {
             .session
             .sync_actions(&[xr::ActiveActionSet::new(&info_set)])
             .unwrap();
+        self.interaction_profile_changed(session_data);
```
prikhi, line 174: Initially had this in `openxr_data.restart_session` after the call to this function, but this seemed like a more fitting location.

SpookySkeletons:
Are you sure this is not handled by https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2722

prikhi:
> Are you sure this is not handled by https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2722

Dope, that commit w/ xrizer `main` fixes hand positions on startup as well, thanks!

PR #293, opened 2026-01-31
[#255] Allow Haptic Pulses When Using Non-Legacy Actions

prikhi:
Modify the `input/legacy` module's `legacy_haptic` to send a vibration pulse when using the action manifest input system instead of the legacy input system.

Previously, we would simply exit early if encountering a `ManifestLoadedActions`. Now, we try to find the haptic action in the manifest & call `apply_feedback` on it with a full amplitude vibration for the desired duration.

This fixes an issue in Audio Trip, which provides an action manifest, but uses the legacy `TriggerHapticPulse` interface function instead of the `TriggerHapticVibrationAction` interface function.

Refs #255

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1025,7 +1025,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
         };
 
         let set_map = self.set_map.read().unwrap();
-        let mut sync_sets = Vec::with_capacity(active_sets.len() + 1);
+        let mut sync_sets = Vec::with_capacity(active_sets.len() + 3);
         {
             tracy_span!("UpdateActionState generate active sets");
             for set in active_sets {
@@ -1044,6 +1044,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
                 &data.input_data.pose_data.get().unwrap().set,
             ));
             sync_sets.push(xr::ActiveActionSet::new(&skeletal_input.set));
+            sync_sets.push(xr::ActiveActionSet::new(&actions.haptic_set));
             self.legacy_state.on_action_sync();
         }
 
@@ -1562,6 +1563,8 @@ struct ManifestLoadedActions {
     per_profile_bindings: HashMap<xr::Path, SecondaryMap<ActionKey, Vec<BindingData>>>,
     info_set: xr::ActionSet,
     _info_action: xr::Action<bool>,
+    haptic_set: xr::ActionSet,
+    haptic_action: xr::Action<xr::Haptic>,
 }
 
 impl ManifestLoadedActions {
--- a/src/input/action_manifest.rs
+++ b/src/input/action_manifest.rs
@@ -133,12 +133,27 @@ impl<C: openxr_data::Compositor> Input<C> {
         let info_action = info_set
             .create_action::<bool>("xrizer-info-action", "XRizer info action", &[])
             .unwrap();
+        // Generate an action set & action for handling haptic pulses.
+        // See `System::TriggerHapticPulse` & `Input::legacy_haptic`.
+        let haptic_set = self
+            .openxr
+            .instance
+            .create_action_set("xrizer-haptic-set", "XRizer haptic set", 0)
+            .unwrap();
+        let haptic_action = haptic_set
+            .create_action::<xr::Haptic>(
+                "xrizer-haptic-action",
+                "XRizer haptic action",
+                &[self.subaction_paths.left, self.subaction_paths.right],
+            )
+            .unwrap();
 
         let mut binding_context = BindingsLoadContext::new(
             &sets,
             actions,
             &session_data.input_data.pose_data.get().unwrap().grip,
             &info_action,
+            &haptic_action,
             skeletal_input,
         );
 
@@ -161,6 +176,7 @@ impl<C: openxr_data::Compositor> Input<C> {
             .chain([
                 &session_data.input_data.pose_data.get().unwrap().set,
                 &info_set,
+                &haptic_set,
                 &skeletal_input.set,
             ])
             .collect();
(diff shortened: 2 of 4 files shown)


src/input/legacy.rs
```diff
@@ -92,20 +92,18 @@ impl<C: openxr_data::Compositor> Input<C> {
...
-        let hand_path = match hand {
-            Hand::Left => self.subaction_paths.left,
-            Hand::Right => self.subaction_paths.right,
-        };
+        let hand_path = self.get_subaction_path(hand);
+
+        let data = self.openxr.session_data.get();
+        if let Some(manifest_actions) = data.input_data.get_loaded_actions() {
+            // Game provided action manifest but also calls the legacy action's pulse method.
+            self.legacy_haptic_via_manifest(manifest_actions, hand_path, duration_us);
```
prikhi, line 104: We could implement this in `input.rs` & have `System.TriggerHapticPulse` defer to the correct function instead of doing that here, but that relies on us exposing some way to determine if we are in legacy or manifest input system, which is not currently exposed in the modules. Could do it, but wanted to prob for feedback first.

src/input/legacy.rs
```diff
@@ -132,6 +130,41 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+    ) {
+        trace!("triggered legacy haptic while using action manifest");
+        let Some(haptic_action) =
+            manifest_actions
+                .actions
+                .iter()
+                .find_map(|(_k, action)| match action {
+                    ActionData::Haptic(haptic_action) => Some(haptic_action),
+                    _ => None,
+                })
```
prikhi: is there a simpler way to get the haptic action?

Supreeeme: I think it would make more sense to just always have a haptic action that we create and control when we have a manifest, instead of relying on the app to provide a haptic action.

prikhi: hmm, that would be nice, esp as i have a additional (unpushed) commit that sets up default haptics if a manifest's actionset has a vibration event but the binding files have no `haptics` key for that actionset - purely to fix audio trip 🤔

prikhi: alright, poked around a bit - I see we're already doing something similar with `info_set` & `info_action` in `action_manifest`. Tossing in another "haptic" action set / action sound good?

prikhi:
@Supreeeme second commit here adds a top-level haptic actionset & action to the ManifestLoadedActions, loads it during UpdateActionState calls, and uses it in the HapticPulse interface func if we have a manifest instead of legacy bindings.

Verified as working in Audio Trip & closes #255

src/input.rs
```diff
@@ -1025,7 +1025,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput010_Interface for Input<C> {
         };
 
         let set_map = self.set_map.read().unwrap();
-        let mut sync_sets = Vec::with_capacity(active_sets.len() + 1);
+        let mut sync_sets = Vec::with_capacity(active_sets.len() + 3);
```
prikhi, line 1028: 1 was wrong as previous value, there were `n` active_sets, the pose_data set, and the skeletal_input set. W/ the haptic_set, there are now 3.

src/input.rs
```diff
@@ -1562,6 +1563,8 @@ struct ManifestLoadedActions {
     per_profile_bindings: HashMap<xr::Path, SecondaryMap<ActionKey, Vec<BindingData>>>,
     info_set: xr::ActionSet,
     _info_action: xr::Action<bool>,
+    haptic_set: xr::ActionSet,
+    haptic_action: xr::Action<xr::Haptic>,
```
prikhi, line 1567: can toss a doc comment on these if ya want

src/input/legacy.rs
```diff
@@ -132,6 +130,31 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+    /// legacy-input haptic interface functions while providing manifest files.
+    fn legacy_haptic_via_manifest(
+        &self,
+        manifest_actions: &ManifestLoadedActions,
+        hand_path: xr::Path,
+        duration_us: ::std::ffi::c_ushort,
+    ) {
+        trace!("triggered legacy haptic while using action manifest");
+        manifest_actions
+            .haptic_action
```
prikhi, line 146: now uses the loaded manifest's global action

src/input/action_manifest.rs
```diff
@@ -133,12 +133,27 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+        // See `System::TriggerHapticPulse` & `Input::legacy_haptic`.
+        let haptic_set = self
+            .openxr
+            .instance
+            .create_action_set("xrizer-haptic-set", "XRizer haptic set", 0)
+            .unwrap();
+        let haptic_action = haptic_set
+            .create_action::<xr::Haptic>(
+                "xrizer-haptic-action",
+                "XRizer haptic action",
```
prikhi, line 146: dunno if the set/action naming is important but just tried to match the info set/action from above.

src/input/action_manifest.rs
```diff
@@ -169,7 +185,10 @@ impl<C: openxr_data::Compositor> Input<C> {
         // Try forcing an interaction profile now
         session_data
             .session
-            .sync_actions(&[xr::ActiveActionSet::new(&info_set)])
+            .sync_actions(&[
+                xr::ActiveActionSet::new(&info_set),
+                xr::ActiveActionSet::new(&haptic_set),
```
Supreeeme: doesn't need to be added here

src/input/legacy.rs
Supreeeme: I don't see a reason to do that

src/input/action_manifest.rs
prikhi: word, in the middle of playing this right now so i'll test it out rq :sweat_smile:

prikhi: yup yup, fixed in latest

src/input/legacy.rs
prikhi: only reason was seems "off" to be doing manifest stuff in the legacy module, but if you're good w/ it, then sounds good to me

prikhi:
thanks for merge, your assistance, & your patience!

PR #297, opened 2026-02-05
openxr_data: improve error message when app is using wrong GPU

ImSapphire:


src/openxr_data.rs
```diff
@@ -430,9 +431,34 @@ impl SessionData {
...
+                            vk::Instance::from_raw(info.instance as _),
+                        )
+                    };
+                    let get_pd_name = |pd| {
+                        let pd_properties = unsafe { inst.get_physical_device_properties(pd) };
+                        pd_properties
+                            .device_name_as_c_str()
+                            .unwrap()
+                            .to_string_lossy()
+                            .to_string()
```
```suggestion
                        pd_properties
                            .device_name_as_c_str()
                            .map(|name| {
                                name.to_string_lossy()
                                .to_string()
                            })
                            .unwrap_or_else(|| "<unknown>".to_string())
```

src/openxr_data.rs
```diff
@@ -430,9 +431,34 @@ impl SessionData {
...
+                );
+                let pd = vk::PhysicalDevice::from_raw(info.physical_device as _);
+
+                if expected_pd != pd {
+                    let inst = unsafe {
+                        ash::Instance::load(
+                            ash::Entry::load().unwrap().static_fn(),
+                            vk::Instance::from_raw(info.instance as _),
+                        )
+                    };
```
Supreeeme, line 446: Might be worth also just finding a way to pass through the vulkan instance to this function.

ImSapphire: I can't really figure out a good way to do this because of how many abstractions the graphics backend and session restart code goes through. it seems like the `VulkanData` with the `ash::Instance` is dropped, so not sure that it's possible as-is?

Supreeeme: You should be able to pull it off the `DynFrameController` on the `Compositor`, with a simple `match`.

PR #298, opened 2026-02-06
input: add Vive Focus3 controller profile

ImSapphire:


src/input/profiles/vive_focus3.rs
```diff
@@ -0,0 +1,260 @@
...
+            PathTranslation {
+                from: "a/touch",
+                to: "a/click",
+                stop: true,
+            },
+            PathTranslation {
+                from: "b/touch",
+                to: "b/click",
+                stop: true,
+            },
```
Supreeeme, line 72: Is this right? This seems like it would be problematic if something that's supposed to be a touch binding is mapped as a click.

src/input/profiles.rs
```diff
@@ -114,6 +116,7 @@ impl Profiles {
                 (ControllerType::Knuckles, &Knuckles),
                 (ControllerType::OculusTouch, &Touch),
                 (ControllerType::ViveController, &SimpleController),
+                (ControllerType::ViveFocus3, &ViveFocus3),
```
Supreeeme, line 120: Similar to #130, there will need to be some sort of check for this extension before we suggest bindings for it.

src/input/profiles/vive_focus3.rs
ImSapphire: The Focus 3 interaction profile doesn't have any touch inputs: https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_HTC_vive_focus3_controller_interaction
which is weird, since I'm pretty sure on SteamVR it's supported..

Supreeeme:
Looks like the test will need to enable the extension. Should be simple enough to add to fakexr.

ImSapphire:
Oops, should be good now

Issue #246, opened 2025-12-09
[Feature]: 3D model of controller on VR program


GK-Gaming:
Add passing the 3D model of the controller used to the VR program so that it shows up correctly there.
And to show the controller in loading screens.

RinLovesYou:
There's a repo with controller models over [here](https://github.com/immersive-web/webxr-input-profiles/tree/main/packages/assets/profiles), that is permissively licensed, which includes full controller models with seperated parts that could even be animated.

PR #302, opened 2026-02-10
render models

galister:
PoC for render models. related ticket: https://github.com/Supreeeme/xrizer/issues/246

~~[temporary] Set `STEAMVR_ROOT` to steamapps/common/SteamVR and it will load the models from there.~~

I did mess with loading components from JSON but then it seemed like too much effort so I removed most of that. Static models are much simpler and they're already a huge upgrade from no models.

Ideally I would ditch the JSONs and would just distribute static models with the correct offsets baked in.

After investigating some of the models, the ones that ship with SteamVR seem to be identical in topology to the ones in this permissively licensed project here: https://github.com/immersive-web/webxr-input-profiles/tree/main/packages/assets/profiles

Not sure what would be the best solution for distributing the models and all. Comments welcome.

### DIFF
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -950,7 +950,8 @@ impl vr::IVROverlay028_Interface for OverlayMan {
         _: u32,
         _: u32,
     ) -> vr::EVROverlayError {
-        todo!()
+        crate::warn_unimplemented!("SetOverlayIntersectionMask");
+        vr::EVROverlayError::None
     }
     fn IsHoverTargetOverlay(&self, _: vr::VROverlayHandle_t) -> bool {
         todo!()


GK-Gaming:
doesnt WiVRn also render models in its menu by itself, even with animated components?
is it already possible to pass these with OpenXR or libmonado?
at least its way of loading and animating the models could be inspire code here maybe?
sorry if this is a bother, just trying to pass some ideas and sorry if these ideas seemed obvious already idk...

galister:
I got the assets down to around 20-50kb per render model by using compressed GLB & omitting textures.

If needed, the sizes can be further reduced by:
- Decimating geometry
- Omitting UVs (there's not really a downside, unless we plan to use textures later on)
- Omitting normals & calculating them at import time instead

For now, I only plan to include models that are actually referenced via the `InteractionProfile` trait.

@Supreeeme not sure if it's preferred to ship a resources folder, or just `include_bytes!` the assets.

galister:
Decided to `include_bytes` the models. With this, we're looking at <200kB added to the binary.

The models are low-poly, they are meant to show the silhouette of the controller and have no features such as buttons. UVs and normals have also been stripped.

I really wanted to do GLB with draco compression, but rust support is not great on that front, so lz4-compressed low-poly obj files it is for now. Once draco is available, we can technically switch to high-poly models while maintaining similar file sizes.

```
2.6K  generic_controller.obj.lz4
 36K  oculus_quest2_controller_left.obj.lz4
 36K  oculus_quest2_controller_right.obj.lz4
 35K  valve_controller_knu_1_0_left.obj.lz4
 30K  valve_controller_knu_1_0_right.obj.lz4
 21K  vr_controller_vive_1_5.obj.lz4
 16K  vr_tracker_vive_3_0.obj.lz4
 ```

File: src/rendermodels.rs [hunk 1]
```diff
@@ -1,13 +1,97 @@
...
+        let mid = lo + (hi - lo) / 2;
+        let mid_key = RENDER_MODELS[mid].0;
+
+        match mid_key.cmp(model_name) {
+            Ordering::Less => lo = mid + 1,
+            Ordering::Greater => hi = mid,
+            Ordering::Equal => return Some(RENDER_MODELS[mid].1),
+        }
+    }
+    None
```
prikhi: Use [`binary_search_by_key`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by_key)?

src/rendermodels.rs
```diff
@@ -1,13 +1,97 @@
...
+use glam::Vec3;
+use log::{debug, warn};
+use lz4_flex::frame::FrameDecoder;
+use obj::{IndexTuple, ObjData};
 use openvr as vr;
 
+// use a single static component for all rendermodels
+static DEFAULT_COMPONENT: &CStr = c"xrizer_component";
+
+// must be in strict alphabetical order
```
prikhi: Have a test that asserts this?

src/rendermodels.rs
```diff
@@ -38,42 +122,88 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+        let render_model_name = unsafe { CStr::from_ptr(render_model_name) };
+        let component_name = unsafe { CStr::from_ptr(component_name) };
+
+        if component_name != DEFAULT_COMPONENT || get_render_model_data(render_model_name).is_none()
+        {
+            // unsupported model or component
+            return false;
+        }
+
+        // all of our models are static and have offsets baked in
```
prikhi: don't know much about this but will the "baked in offsets" be wrong for clients that desire different pose types(raw, base, tip, etc) if/when we get to implementing controller pose transforms?

src/rendermodels.rs
galister: done, thanks

src/rendermodels.rs
galister: good idea, added.

src/rendermodels.rs
galister: base / tip / etc poses are defined relative to raw in the same render model json that the rendermodel component is defined. 

in our implementation, we make it so that the model is already aligned, and we tell the openvr client to render it at identity transform (this is done in GetComponentState), meaning the openvr client will render it at raw pose with no offset. which is what we want, as the offsets are in the model already.

galister:
apparently left-right controllers all seem to be just mirrored on the global x axis.

i will look into only embedding left ones and calculating the right mesh at runtime

galister:
for oculus touch and index, i've changed to embedding only left and calculating right at load time. the old method is still available for controllers added in the future where this doesn't apply.

with this, the current size of all models in total is 110kB.

galister:
rebased for vive focus 3 controllers

src/rendermodels.rs
```diff
@@ -1,13 +1,151 @@
...
 
+// use a single static component for all rendermodels
+static DEFAULT_COMPONENT: &CStr = c"xrizer_component";
+
+struct RenderModelData {
+    bytes: &'static [u8],
+    flipped: bool,
+}
+
+const RENDER_MODELS: &[(&str, RenderModelData)] = {
```
Supreeeme: I think this should be `static`, `const` would be the equivalent of copy-pasting this everywhere.

src/rendermodels.rs
```diff
@@ -1,13 +1,151 @@
...
+    let oculus_quest2_controller_left =
+        include_bytes!("../resources/rendermodels/oculus_quest2_controller_left.obj.lz4");
+    let oculus_quest_plus_controller_left =
+        include_bytes!("../resources/rendermodels/oculus_quest_plus_controller_left.obj.lz4");
+    let valve_controller_knu_1_0_left =
+        include_bytes!("../resources/rendermodels/valve_controller_knu_1_0_left.obj.lz4");
+    let vive_focus3_controller_left =
+        include_bytes!("../resources/rendermodels/vive_focus3_controller_left.obj.lz4");
+
+    // must be in strict alphabetical order
```
Supreeeme: Why?

src/rendermodels.rs
```diff
@@ -1,13 +1,151 @@
...
 #[derive(Default, macros::InterfaceImpl)]
 #[interface = "IVRRenderModels"]
 #[versions(006, 005, 004)]
 pub struct RenderModels {
     vtables: Vtables,
+    data: Mutex<RenderModelMutex>,
+}
+
+#[derive(Default)]
+struct RenderModelMutex {
```
Supreeeme:
```suggestion
struct RenderModelData {
```

src/rendermodels.rs
```diff
@@ -38,42 +176,105 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+
+        if component_name != DEFAULT_COMPONENT || get_render_model_data(render_model_name).is_none()
+        {
+            // unsupported model or component
+            return false;
+        }
+
+        // all of our models are static and have offsets baked in
+        unsafe {
+            std::ptr::write_bytes(state, 0, 1);
```
Supreeeme: Why? We write all the fields already.

src/rendermodels.rs
```diff
@@ -38,42 +176,105 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+            (*state).mTrackingToComponentRenderModel = vr::HmdMatrix34_t {
+                m: [[1., 0., 0., 0.], [0., 1., 0., 0.], [0., 0., 1., 0.]],
+            };
+            (*state).mTrackingToComponentLocal = vr::HmdMatrix34_t {
+                m: [[1., 0., 0., 0.], [0., 1., 0., 0.], [0., 0., 1., 0.]],
+            };
```
Supreeeme: The `openvr` crate has a implementation of `From<xr::Posef>` for `HmdMatrix34_t` and I think using that here would be clearer.

src/rendermodels.rs
```diff
@@ -91,40 +292,42 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
     }
     fn GetComponentCount(&self, render_model_name: *const std::ffi::c_char) -> u32 {
-        let name = unsafe { CStr::from_ptr(render_model_name) };
-        debug!("getting components for {name:?}");
+        if render_model_name.is_null() {
+            return 0;
+        }
 
-        if name.count_bytes() == 0 { 0 } else { 1 }
+        1 // DEFAULT_COMPONENT
```
Supreeeme: There's no check for the component count anymore.

src/rendermodels.rs
```diff
@@ -134,34 +337,810 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+struct OwnedTexture {
+    width: u16,
+    height: u16,
+    data: Box<[u8]>,
+}
+
+impl OwnedTexture {
+    /// creates a 1x1 dummy texture
+    fn new_dummy() -> Self {
+        let data = Box::new([32, 32, 32, 255]);
```
Supreeeme: Are these numbers arbitrary?

src/rendermodels.rs
```diff
@@ -134,34 +337,810 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+            unWidth: self.width,
+            unHeight: self.height,
+            unMipLevels: 1,
+            rubTextureMapData: self.data.as_ptr(),
+            format: openvr::EVRRenderModelTextureFormat::RGBA8_SRGB,
+        })
+    }
+}
+
+#[allow(dead_code)]
```
Supreeeme: Why is this dead code?

src/rendermodels.rs
```diff
@@ -134,34 +337,810 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+struct OwnedRenderModel {
+    verts: Box<[vr::RenderModel_Vertex_t]>,
+    indices: Box<[u16]>,
+}
+
+impl OwnedRenderModel {
+    fn load(name: &CStr, model_data: &RenderModelData) -> Result<Self, vr::EVRRenderModelError> {
+        let mut decompressed_bytes = vec![];
+        FrameDecoder::new(model_data.bytes)
```
Supreeeme: `RenderModelData` could use a comment clarifying that `bytes` is lz4 compressed data.

src/rendermodels.rs
galister: adjusted, thanks.

src/rendermodels.rs
galister: we use slice::binary_search_by_key for lookup, so it needs to be pre-sorted. there's also a test to enforce the ordering

src/rendermodels.rs
galister: adjusted.

src/rendermodels.rs
galister: adjusted.

src/rendermodels.rs
galister: made it into a const instead.

src/rendermodels.rs
galister: adjusted.

src/rendermodels.rs
galister: oops.

src/rendermodels.rs
galister: also renamed the struct to be more straightforward.

galister:
hi, anything else in mind for this?

src/rendermodels.rs
```diff
@@ -134,34 +344,817 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+
+        for object in &obj.objects {
+            for group in &object.groups {
+                for poly in &group.polys {
+                    let poly_indices = &poly.0;
+
+                    if poly_indices.len() < 3 {
+                        continue;
+                    }
+
```
Supreeeme: Is a warning needed here?

src/rendermodels.rs
```diff
@@ -134,34 +344,817 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+                        if flipped {
+                            // reverse winding to keep normals/front-faces correct
+                            indices.push(i0);
+                            indices.push(i2);
+                            indices.push(i1);
+                        } else {
+                            indices.push(i0);
+                            indices.push(i1);
+                            indices.push(i2);
+                        }
```
Supreeeme:
```suggestion
                        if flipped {
                            // reverse winding to keep normals/front-faces correct
                            indices.extend([i0, i2, i1]);
                        } else {
                            indices.extend([i0, i1, i2]);
                        }
```

src/rendermodels.rs
```diff
@@ -134,34 +344,817 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+                    }
+                }
+            }
+        }
+
+        // calculate normals
+        let mut accum = vec![Vec3::ZERO; verts.len()];
```
Supreeeme:
```suggestion
        let mut normals = vec![Vec3::ZERO; verts.len()];
```
I assume these are supposed to be normals?

Supreeeme (commented):
The rendermodels folder could use some kind of file explaining what they are and where they came from

src/rendermodels.rs
Supreeeme: Couldn't this just be a HashMap?

galister: idk, do you want it to be a hashmap? for me it seems like too much fuss (LazyLock) for something that can just be a static slice with a comment.

src/rendermodels.rs
galister: not really, but added it anyway.

src/rendermodels.rs
galister: yup.

src/rendermodels.rs
galister: added manually.

src/rendermodels.rs
Supreeeme: I don't really see how LazyLock is fuss, it's pretty simple, and wouldn't require a comment and a test.

galister: I took another look at HashMap and remembered why I gave up on it in the first place. The interface methods not only access render models by name, but also by index. Once you take into account that it needs to support lookup by both name and index, I think it's cleaner this way.

As for the test, we can actually remove the alphabetic order test as 
`get_render_model_data_able_to_resolve_all_elements` also covers the same (and is needed for coverage of the supporting logic in `get_render_model_data`).
I just think the alphabetical test gives a more helpful message in case someone screws it up.

galister:
Hi, anything else I can do for this one?

src/rendermodels.rs
```diff
@@ -38,42 +181,103 @@ impl vr::IVRRenderModels006_Interface for RenderModels {
...
+        // all of our models are static and have offsets baked in
+        let identity = xr::Posef::IDENTITY;
+        let identity = vr::HmdMatrix34_t::from(identity);
+
+        unsafe {
+            (*state).mTrackingToComponentRenderModel = identity;
+            (*state).mTrackingToComponentLocal = identity;
+            (*state).uProperties =
+                (vr::EVRComponentProperty::IsVisible | vr::EVRComponentProperty::IsStatic).0;
+        }
```
Supreeeme:
```suggestion
        unsafe {
            (*state).mTrackingToComponentRenderModel = xr::Posef::IDENTITY.into();
            (*state).mTrackingToComponentLocal = xr::Posef::IDENTITY.into();
            (*state).uProperties =
                (vr::EVRComponentProperty::IsVisible | vr::EVRComponentProperty::IsStatic).0;
        }
```

src/rendermodels.rs
Supreeeme: I don't think the Vec is necessarily cleaner, because we have to guarantee that it's sorted. Not that this is a particularly challenging requirement to uphold, but it just feels too implicit and runtime for me, and I fear it getting lost down the line. Given that the majority of render model access is through the render model name (in terms of API surface), and the fact that the map of render models never changes, it makes the most sense to me to just have it be a HashMap and for `GetRenderModelName` to just use `RENDER_MODELS.iter().nth(index)` to grab a render model by index. (It doesn't say it in the docs, but I think the iteration order of the HashMap shouldn't change if the map itself doesn't change. But if you are concerned with that then it could also just be a BTreeMap.)

PR #306, opened 2026-02-12
[#305] Add OpenVR v0.9.19 for BoneWorks

### LINKED ISSUE #305: [Bug]: App (Boneworks) Requested IVRCompositor_013
### Operating System

Arch Linux

### OpenXR Runtime

Monado

### Describe the issue.

I have a Vive Pro 1 headset and two index knuckles controllers. When testing Boneworks (from steam), haptics on the knuckles controllers seem not to be implemented. Correct haptics can be felt when using SteamVR. 

I received some great help on Discord, and I was pointed to this version of openvr:
https://github.com/ValveSoftware/openvr/blob/v0.9.19/headers/openvr.h
which is the latest version that has the compositor mentioned in the log I uploaded.

[xrizer.txt](https://github.com/user-attachments/files/25237436/xrizer.txt)

This may not be strictly relevant, but I also tested Boneworks with the ImSapphire fork of Xrizer and had complete finger tracking support, though still a lack of haptics. Lack of finger tracking support is documented here: https://github.com/Supreeeme/xrizer/issues/251

Please let me know if you need any more information or I can do anything else to help.

### Steps to reproduce

I am using the Envision software, using latest commit of both Monado and Xrizer.

Step 1: Launch Monado+Xrizer through Envision
Step 2: Launch Boneworks through Steam with launch option: PRESSURE_VESSEL_IMPORT_OPENXR_1_RUNTIMES=1 %command%
Step 3: Observe lack of haptic feedback when grabbing objects/hitting hands against geometry

prikhi:
BoneWorks requires `IVRCompositor_013` & v0.9.19 is the latest version that supports it.

Additionally, pulled in `IVRSettings_001`, since that is defined in v0.9.19 as well.

Closes #305 but don't have game so will need verification from issue submitter.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -6,7 +6,7 @@ use std::os::raw::c_char;
 
 #[derive(Default, macros::InterfaceImpl)]
 #[interface = "IVRSettings"]
-#[versions(003, 002)]
+#[versions(003, 002, 001)]
 pub struct Settings {
     vtables: Vtables,
 }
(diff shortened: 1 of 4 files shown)


src/settings.rs
```diff
@@ -185,3 +185,54 @@ impl vr::IVRSettings002On003 for Settings {
...
+
+impl vr::IVRSettings001On002 for Settings {
+    fn GetBool(
+        &self,
+        section: *const c_char,
+        settings_key: *const c_char,
+        _default_value: bool,
+        error: *mut EVRSettingsError,
+    ) -> bool {
+        vr::IVRSettings003_Interface::GetBool(self, section, settings_key, error)
```
prikhi, line 197: These just call the latest interface functions, which don't take the `default_value` params. They just return constant values(e.g., `false`) instead of actually storing/fetching settings. 

Is that weird? Should we have like a `HashMap<String, HashMap<String, SettingsValue>>` that we utilize as we receive `Set` & `Get` calls? Does anything ever care about these settings?

prikhi (commented):
Yet another `Compositor_FrameTiming`. And dumb stubs for the new settings interface included in this version since we don't actually track/modify settings values. All other interface versions found in `.19` either don't exist in the code at all or have been implemented already from past/future versions.

prikhi:
@Supreeeme this wasn't actually needed to get the game to work(oddly it requests 2 diff versions of ivrcompositor & is fine running w/ one of the versions we already implement), so up to you if you wanna merge or just close this out.

PR #308, opened 2026-02-14
input: add Touch Plus profile

ImSapphire:


Supreeeme:
I'm working on a refactor of how input profiles are implemented, so I'm going to wait to merge this until I'm done with that.

Supreeeme:
Took longer than expected, but the refactor is now done.

ImSapphire:
Rebased

Supreeeme:
Is there a fundamental reason for this profile to be added? It seems identical to the normal touch.

skryvel:
I'd been working on something similar to reach a happy Quest 3 experience. I started going [in this direction before noticing this PR here](https://github.com/Supreeeme/xrizer/compare/main...skryvel:xrizer-fork:quest3-controlers).

> identical to the normal touch

Not sure if this is a relevant answer but,  `origin/main` as of now does not work (completely at least) with Quest 3 controllers.

ImSapphire:
> Is there a fundamental reason for this profile to be added? It seems identical to the normal touch.

It's mainly for the rendermodel. On main it uses the Quest 2 rendermodel.

Supreeeme:
In that case, I would derive as much of the functions as possible from the existing Touch profile.

ImSapphire:
> In that case, I would derive as much of the functions as possible from the existing Touch profile.

Do you mean like, replacing the body of e.g. legacy_bindings with `OculusTouch::legacy_bindings(c)`? I can't do that because of the generic type.

Supreeeme:
You could create a trait, `OculusLike` or whatever - have the stuff that needs to be unique between the profiles as required trait functions, and add a blanket implementation for all types that implement `OculusLike`. There are quite a few touch-like controllers anyway like the Pico controllers, so that could end up pretty useful I think.

PR #311, opened 2026-02-14
openxr_data: use now as initial display time

ImSapphire:
Apps may call GetProjectionRaw, GetProjectionMatrix, GetEyeToHeadTransform, GetHiddenAreaMesh, etc. before calling WaitGetPoses, and the runtime may not give us good views when requesting at a time of 1.
    
hellovr_opengl for example calls GetEyeToHeadTransform and GetProjection{Raw,Matrix} on startup, if the runtime doesn't give valid views the camera perspective is completely wrong.


Supreeeme:
The PR description should also go in the commit description.

ImSapphire:
> The PR description should also go in the commit description.

done

PR #314, opened 2026-02-17
system, compositor: report actual refresh rate

ImSapphire:


src/compositor.rs
```diff
@@ -622,7 +626,10 @@ impl vr::IVRCompositor029_Interface for Compositor {
             set!(m_flCompositorRenderCpuMs, 3.0);
             set!(m_flCompositorIdleCpuMs, 0.1);
 
-            set!(m_flClientFrameIntervalMs, 11.1);
+            set!(
+                m_flClientFrameIntervalMs,
+                1000.0 / self.openxr.get_refresh_rate()
```
Supreeeme, line 631: Not obvious why this is being divided by 1000.

src/openxr_data.rs
```diff
@@ -173,6 +175,7 @@ impl<C: Compositor> OpenXrData<C> {
             system_id,
             session_data,
             display_time: AtomicXrTime(1.into()),
+            display_period: 11111111.into(),
```
Supreeeme: Why?

src/openxr_data.rs
```diff
@@ -308,6 +311,20 @@ impl<C: Compositor> OpenXrData<C> {
         };
     }
 
+    pub fn get_refresh_rate(&self) -> f32 {
+        let get_fallback_rate =
+            || (1_000_000_000f64 / self.display_period.load(Ordering::Relaxed) as f64) as f32;
```
```suggestion
            || std::time::Duration::from_nanos(self.display_period.load(Ordering::Relaxed)).as_secs_f32();
```

src/compositor.rs
ImSapphire: It's to convert the refresh rate to the interval in milliseconds, eg. `1000.0 / 90.0 = 11.111111...`

src/openxr_data.rs
ImSapphire: It's just a placeholder, like the display_time, since it gets overwritten on the first frame anyway

Supreeeme: Would appreciate a comment for this.

src/compositor.rs
Supreeeme, line 631:
```suggestion
                Duration::from_nanos(self.openxr.get_refresh_rate()).as_millis_f32()
```

src/openxr_data.rs
```diff
@@ -40,6 +41,7 @@ pub struct OpenXrData<C: Compositor> {
     pub system_id: xr::SystemId,
     pub session_data: SessionReadGuard,
     pub display_time: AtomicXrTime,
+    pub display_period: AtomicI64,
```
Supreeeme: Need a comment clarifying that this is nanoseconds.

src/compositor.rs
ImSapphire: get_refresh_rate returns FPS as float (like 90.0) so this doesn't work

ImSapphire:
> Is this fixing something, or just for accuracy?

it's just for accuracy, the VRCMark world in VRChat uses the DisplayFrequency property in the results to display the headset refresh rate

src/openxr_data.rs
```diff
@@ -40,6 +41,8 @@ pub struct OpenXrData<C: Compositor> {
     pub system_id: xr::SystemId,
     pub session_data: SessionReadGuard,
     pub display_time: AtomicXrTime,
+    /// in ns
+    pub display_period: AtomicI64,
```
```suggestion
    pub display_period_nanos: AtomicI64,
```

Supreeeme (commented):
final nit

src/openxr_data.rs
ImSapphire: done

PR #315, opened 2026-02-19
Support System Button in Knuckles Bindings

prikhi:
Modify the `profiles::knuckles` module to allow binding of actions to `system/touch` & `system/click` for both hands. No transformation is necessary as the path mapping is identical between the OpenXR spec & SteamVR.

Note that OpenXR runtimes _may_ choose to ignore bindings to the system buttons.

### DIFF
--- a/src/input/profiles/knuckles.rs
+++ b/src/input/profiles/knuckles.rs
@@ -89,9 +89,15 @@ impl InteractionProfile for Knuckles {
     }
 
     fn legal_paths(&self) -> Box<[String]> {
-        let click_and_touch = ["input/a", "input/b", "input/trigger", "input/thumbstick"]
-            .iter()
-            .flat_map(|p| [format!("{p}/click"), format!("{p}/touch")]);
+        let click_and_touch = [
+            "input/a",
+            "input/b",
+            "input/trigger",
+            "input/thumbstick",
+            "input/system",
+        ]
+        .iter()
+        .flat_map(|p| [format!("{p}/click"), format!("{p}/touch")]);
         let x_and_y = ["input/thumbstick", "input/trackpad"]
             .iter()
             .flat_map(|p| [format!("{p}/x"), format!("{p}/y"), p.to_string()]);
@@ -199,6 +205,8 @@ mod tests {
                 "/user/hand/right/input/a/click".into(),
                 "/user/hand/left/input/b/click".into(),
                 "/user/hand/right/input/b/click".into(),
+                "/user/hand/left/input/system/click".into(),
+                "/user/hand/right/input/system/click".into(),
                 "/user/hand/left/input/trigger/touch".into(),
                 "/user/hand/right/input/trigger/touch".into(),
                 "/user/hand/left/input/thumbstick/click".into(),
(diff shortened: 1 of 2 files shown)


src/input/profiles/knuckles.rs
```diff
@@ -89,9 +89,15 @@ impl InteractionProfile for Knuckles {
...
     fn legal_paths(&self) -> Box<[String]> {
-        let click_and_touch = ["input/a", "input/b", "input/trigger", "input/thumbstick"]
-            .iter()
-            .flat_map(|p| [format!("{p}/click"), format!("{p}/touch")]);
+        let click_and_touch = [
+            "input/a",
+            "input/b",
+            "input/trigger",
+            "input/thumbstick",
+            "input/system",
```
prikhi, line 97: click & touch for both hands as per openxr spec: https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#_valve_index_controller_profile

and as in steamvr drivers:
```jsonc
// driver/.../index_controller_profile.json
  "input_source" :
  {
    "/input/system" : {
      "type" : "button",
      "click" : true,
      "touch" : true,
      "binding_image_point" : [ 34,45 ],
      "order" : 1
    },
// ...
```

With Moss 2 binding file as an example:

```json
            {
               "inputs" : {
                  "click" : {
                     "output" : "/actions/main/in/ui_menu"
                  }
               },
               "mode" : "button",
               "path" : "/user/hand/left/input/system"
            },

```

tests/input_data/knuckles.json
```diff
@@ -1,273 +1,297 @@
 {
-	"bindings": {
```
prikhi, line 2: ugh... I can add back the tabs if desired... 
(forgot to turn nvim autoformat of json off)

Supreeeme:
The spec says that system buttons may not be available for application use. In practice what this means is that depending on what runtime it's running under, suggesting bindings using the system button may actually error out with `XR_ERROR_PATH_UNSUPPORTED`. For this reason, along with the fact that I'm pretty sure even binding to the system button in SteamVR requires changing some setting, I opted to not use system button bindings. I'm not really against supporting them at all, but this case needs to be handled.

prikhi:
> In practice [...] may actually error out with `XR_ERROR_PATH_UNSUPPORTED`

Ah good to know, the spec says it "**must**" accept it but is free to ignore it so figured it wouldn't cause any problems: 
https://registry.khronos.org/OpenXR/specs/1.1/man/html/xrSuggestInteractionProfileBindings.html

> I'm not really against supporting them at all, but this case needs to be handled.

aight thanks, will poke around to see where we're doing this & if the error can be handled cleanly.

prikhi:
> I'm not really against supporting them at all, but this case needs to be handled.

I remember peeking at this but I think at bind time we lost the string path & just had the XrPath or something like that. So I _could_ catch a binding failure but couldn't tell if it was the system button causing it. But I also couldn't find any monado code that would have thrown when attempting to bind the system button :shrug: 

Might re-investigate at some point w/ the new input/binding system but gonna close this for now.

Issue #316, opened 2026-02-20
[Bug]: RE4 VR Mod - XRizer crash on boot - panicked at src/overlay.rs:1320:9


ManMeme1:
### Operating System

CachyOS 6.19.2-2-cachyos

### OpenXR Runtime

WiVRn

### Describe the issue.

Log attached.

When trying to use the RE4 VR Mod that contains Motion Control fixes, Xrizer will crash as soon as the Capcom logo appears on the desktop, while VR is trying to be initialised.

### Steps to reproduce

Install https://www.nexusmods.com/residentevil42023/mods/5682 for RE4. Steps to install this mod are in the description of the mod page.

Launch the game through WayVR, which will trigger the game to launch in VR.

Xrizer will crash when the Capcom logo appears.

[xrizer.txt](https://github.com/user-attachments/files/25452161/xrizer.txt)

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

prikhi:
@ManMeme1, can you try my branch from PR #317  and tell me if it works? If you use envision:

1. Duplicate your profile
2. Set OpenVR Compat Repo to: `https://github.com/prikhi/xrizer.git`
3. Set OpenVR Compat Branch to: `prikhi/316-re4-vr-mod-support`
4. Click `Save` & perform a clean build.

Send new logs if still unplayable. Won't fix all the errors in your log but might stop xrizer from crashing.

ManMeme1:
@prikhi I don't use Envision since I primarily use only WiVRn, I have compiled this PR locally and the game no longer crashes.

It doesn't load in VR however due to being unable to map inputs. Something I will need to look into myself, the mod contains all the bindings needed I believe just pointing them to XRizer for it to properly initialise.

Thank you for producing a fix so soon!

<img width="312" height="334" alt="Image" src="https://github.com/user-attachments/assets/e905437c-6305-4090-b983-490b53323f2d" />

Mr-Zero88:
the openvr input problem is fixed with this https://github.com/prikhi/xrizer/pull/1

prikhi:
I've merged that ☝️  into my branch so let us know how it goes.

ManMeme1:
@prikhi I have tried again with this latest commit, however the game does not boot up, I have tried both with XRizer and SteamVR, both crash at the CapCom Logo.

With help from @Mr-Zero88 It looks to be a Proton issue. Specifically, this line below from Proton Logging.

[...]

PR #317, opened 2026-02-21
[#316] Resident Evil 4 VR Mod Support

prikhi:
Implement the `GetOverlayErrorNameFromEnum` method for the `OverlayMan` struct & `IVROverlay` interface, as required by the Resident Evil 4 VR mod.

I don't own the game so will require testing from @ManMeme1 if this is the only thing blocking game support.


prikhi:
@Supreeeme: comments in #316 say this is enough to make the game work.

PR #329, opened 2026-03-21
Allow old games requiring raw play space boundaries to run

Eidenz:
Some older games are using raw calibration instead of the proper `Standing` and `Seated` types. It's rare for games to use it but causes a crash of XRizer on launch for them.

Proposition: remove unreachable so the game launches anyway, despite the function not being fully supported.

### DIFF
--- a/src/openxr_data.rs
+++ b/src/openxr_data.rs
@@ -631,7 +631,10 @@ impl SessionData {
         match origin {
             vr::ETrackingUniverseOrigin::Seated => &self.local_space_adjusted,
             vr::ETrackingUniverseOrigin::Standing => &self.stage_space_adjusted,
-            vr::ETrackingUniverseOrigin::RawAndUncalibrated => &self.stage_space_reference,
+            vr::ETrackingUniverseOrigin::RawAndUncalibrated => {
+                crate::warn_unimplemented!("RawAndUncalibrated tracking space");
+                &self.stage_space_reference
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/openxr_data.rs
```diff
@@ -629,7 +631,7 @@ impl SessionData {
         match origin {
             vr::ETrackingUniverseOrigin::Seated => &self.local_space_adjusted,
             vr::ETrackingUniverseOrigin::Standing => &self.stage_space_adjusted,
-            vr::ETrackingUniverseOrigin::RawAndUncalibrated => unreachable!(),
+            vr::ETrackingUniverseOrigin::RawAndUncalibrated => &self.stage_space_reference,
```
prikhi: I think at minimum, we should be calling `warn_unimplemented` before returning the fallback data, so that we have something in logs when someone complains about a game being wrong. E.g.:
https://github.com/Supreeeme/xrizer/blob/fc9b0cc2a1047adc9b7edffd68fa0c288965090a/src/applications.rs#L181-L184

src/openxr_data.rs
```diff
@@ -647,8 +649,8 @@ impl SessionData {
     pub fn current_origin_as_reference_space(&self) -> xr::ReferenceSpaceType {
         match self.current_origin {
             vr::ETrackingUniverseOrigin::Seated => xr::ReferenceSpaceType::LOCAL,
-            vr::ETrackingUniverseOrigin::Standing => xr::ReferenceSpaceType::STAGE,
-            vr::ETrackingUniverseOrigin::RawAndUncalibrated => unreachable!(),
+            vr::ETrackingUniverseOrigin::Standing
+            | vr::ETrackingUniverseOrigin::RawAndUncalibrated => xr::ReferenceSpaceType::STAGE,
```
prikhi: I would keep this split so you can also warn here when we get RawAndUncalibrated values

src/openxr_data.rs
Eidenz: This makes sense, I've made the required changes to warn about the use of the fallback stage. It was already an issue with SteamVR too (support was dropped I believe), but always good practice.

Issue #335, opened 2026-04-03
[Bug]: Radium/2021 Rec room incorrect hand offsets


slice-the-idiot:
### Operating System

EndeavorOS

### OpenXR Runtime

WiVRn

### Describe the issue.

[xrizer.txt](https://github.com/user-attachments/files/26462400/xrizer.txt)

For context: Radium is a private server based on the 2021 version of rec room, and seems to have bad hand offsets, iirc normal Rec Room had a similar issue, but that had a hand offset adjustment system so it wasn't a big issue.

### Steps to reproduce

1. launch radium
2. see your hands

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

slice-the-idiot:
just checked live and turns out it recognizes the correct controller while 2021/radium sees them as vive wands, making this issue similar to [another](https://github.com/Supreeeme/xrizer/issues/254)

392luiva:
yo can i have 100k token

slice-the-idiot:
no

also this is no longer necessary as i no longer trust raidum, it was made by hackers and if it's anything i learned from recent littlebigplanet drama do not trust people who cheat in video games ever

slice-the-idiot:
now that i think about it maybe this could fix other games too? either way better safe than sorry, plus i started playing radium again despite my better judgement

Cart1416:
I have this issue on every Rec Room revival I've tried on my Steam Deck with WiVRn and Quest 2. Radium, Stella, EpicQuest, Mocha, Stella 2017, RRL

PR #337, opened 2026-04-08
Add input profile path support for various controllers

Mr-Zero88:


slice-the-idiot:
it should be added that this would fix an [issue i put up ](https://github.com/Supreeeme/xrizer/issues/335)

PR #338, opened 2026-04-09
input: use XR_EXT_hand_tracking_data_source for skeletal tracking level

ImSapphire:


SpookySkeletons:
This fixes the hack checking of index knuckles profile for determining if handtrack should be full or partial. Allows for Mercury handtrack with synthetic valve index hand poses if enabled in the steamvr_lh driver while monado handles reporting which one is active.

src/input/devices.rs
```diff
@@ -21,7 +22,8 @@ pub enum TrackedDeviceType {
     Controller {
         hand: Hand,
         hand_tracker: Option<xr::HandTracker>,
-        skeleton_cache: Mutex<HashMap<u64, Option<xr::HandJointLocations>>>,
+        skeleton_cache:
+            Mutex<HashMap<u64, Option<(xr::HandJointLocations, xr::HandTrackingDataSourceEXT)>>>,
```
Supreeeme: This is a busy type name. I think it'd be worth the clarity to create a dedicated struct instead of the tuple.

src/input/devices.rs
```diff
@@ -179,11 +181,58 @@ impl TrackedDevice {
...
-            .unwrap_or_default();
-        skeleton_cache.insert(base.as_raw().into_raw(), joints);
-        joints
+        fn get_joints_with_data_source(
+            base: &xr::Space,
+            tracker: &xr::HandTracker,
+            time: xr::Time,
+        ) -> xr::Result<Option<(xr::HandJointLocations, xr::HandTrackingDataSourceEXT)>> {
+            let ext = base.instance().exts().ext_hand_tracking.unwrap();
+            unsafe {
```
Supreeeme: I'd prefer if the actual unsafe portions were the only parts marked with unsafe (within reason)

src/input/devices.rs
```diff
@@ -179,11 +181,58 @@ impl TrackedDevice {
...
+                    MaybeUninit::<[xr::HandJointLocation; xr::HAND_JOINT_COUNT]>::uninit();
+                let mut location_info = xr::sys::HandJointLocationsEXT {
+                    ty: xr::sys::HandJointLocationsEXT::TYPE,
+                    next: &mut source_state as *mut _ as _,
+                    joint_count: xr::HAND_JOINT_COUNT as u32,
+                    joint_locations: locations.as_mut_ptr() as _,
+                    ..{ std::mem::zeroed() }
+                };
+                let r =
+                    (ext.locate_hand_joints)(tracker.as_raw(), &locate_info, &mut location_info);
```
Supreeeme: I assume this is using the raw version directly because of the source state addon. That could use a comment considering there is a safe version of `locate_hand_joints`.

src/input/devices.rs
```diff
@@ -179,11 +181,58 @@ impl TrackedDevice {
             return *skeleton;
         }
 
-        let joints = base
-            .locate_hand_joints(hand_tracker.as_ref()?, xr_data.display_time.get())
-            .unwrap_or_default();
-        skeleton_cache.insert(base.as_raw().into_raw(), joints);
-        joints
+        fn get_joints_with_data_source(
```
Supreeeme: This doesn't really need to be a nested function imo, makes it a bit hard to read.

Supreeeme (commented):
Could use some tests 👀

src/input/devices.rs
ImSapphire: I think the reason I did that was for cleaner error handling, I could extract the logic but I think that would be less readable? Unless you mean it should be a function outside of the `impl` block.

ImSapphire:
I didn't write a unit test yet, I don't really have the energy to do that at the moment as it's very difficult for me to figure out how to cleanly write them

Issue #303, opened 2026-02-11
[Bug]: Dungeons of Eternity - Index Controller Profile Points Hands Backwards


prikhi:
### Operating System

Arch Linux

### OpenXR Runtime

Monado

### Describe the issue.

In #256 & my personal tests, I noticed this line in xrizer logs:

```
ERROR xrizer::input::action_manifest ThreadId(1)] Failed to parse bindings for Knuckles: unknown field parameters, expected path or inputs at line 135 column 8
[...]
INFO xrizer::input ThreadId(2)] /user/hand/left interaction profile changed: /interaction_profiles/oculus/touch_controller
INFO xrizer::input ThreadId(2)] /user/hand/right interaction profile changed: /interaction_profiles/oculus/touch_controller
```

I implemented this in the following commit to ensure knuckles profile/bindings were loaded instead of the oculus touch profile: https://github.com/Supreeeme/xrizer/commit/5ce908caaa6c6b2a1642074a91091aeef2d441ad

I also made a separate branch that layered this commit on top of https://github.com/ImSapphire/xrizer/tree/next, as `GetSkeletalSummaryData` is required for weapon unholstering to work in the game(this issue occurs on both `ImSapphire/next` & this repo's `main` branches).

Loading up the game, I can see the parsing error disappears & the knuckles interaction profile gets loaded instead. However, the controller orientation is incorrect(orientation was working correctly when defaulting to oculus interaction profile). Pointing your hand forward results in the game's hands being pointed backwards.

I poked around the xrizer logs a bit more & noticed the `"base"` pose is unimplemented & defaults back to raw pose data:

```
WARN xrizer::input::action_manifest ThreadId(1)] Unknown pose type: "base"
```

Via https://github.com/Supreeeme/xrizer/blob/be664bb93eebdd82336d00230154bacc6f8b9765/src/input/action_manifest.rs#L597-L604 & https://github.com/Supreeeme/xrizer/blob/be664bb93eebdd82336d00230154bacc6f8b9765/src/input.rs#L817-L823

I implemented the `BoundPoseType::Base`, but that obviously does nothing since we only ever return the raw origin/hand.

I poked around SteamVR Drivers, noticed some controller component & each pose type has a transformation matrix:

```jsonc
{ // SteamVR/drivers/indexcontroller/resources/rendermodels/*/*.json
  //...
  "components": {
       // ...
        "tip": {
            "component_local" : {
                "origin": [0.006, -0.015, 0.02],
[...]

PR #382, opened 2026-07-04
[#303] Add Controller Pose Transformations

prikhi:
Add new values to the `BoundPoseType` enum & make it a public type: Base, Handgrip, Grip, OpenxrHandmodel, OpenxrPinch, OpenxrPoke, OpenxrAim, OpenxrGrip.

Modify the `InteractionProfile` type to include a new `pose_transformation` function that returns optional left & right transformation matrices for each `BoundPoseType` via a new `PoseTransformations` type.

Implement the new `pose_transformation` function for the Index Controller, Vive Focus3 Controller, & Oculus Touch Controller profiles - all other profiles always return a `None`.

Extend the `ProfileData` struct to expose a `pose_transformation` function that returns the transformation matrix for a sepcific pose & hand.

Modify the `get_controller_pose` function to take an optional `BoundPoseType` & use it to transform the raw controller location before returning it.

Modify the `GetPoseActionDataForNextFrame` `IVRInput` interface function to use the `BoundPoseType` when calculating & writing the Pose ActionData. This required modifying the type signature of multiple functions in the callstack from `GetPoseActionDataForNextFrame` to `get_controller_pose`. All other uses of the modified functions simply pass in a pose type of `None`.

Modify `TrackedDevice.get_pose` to ignore the `pose_cache` when its `pose_type` argument is defined. This allows the
`GetPoseActionDataForNextFrame` call in a frame to supply the transformed controller locations when `get_pose` has already been called for a device.

Fixes the issue described in #303 where games that use poses other than `Raw` have the hands/controllers pointing in wrong direction.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -297,7 +297,7 @@ struct BoundPose {
 }
 
 #[derive(Clone, Copy, Debug)]
-enum BoundPoseType {
+pub enum BoundPoseType {
     /// Equivalent to what is returned by WaitGetPoses, this appears to be the same or close to
     /// OpenXR's grip pose in the same position as the aim pose.
     /// "All tracked devices also get two pose components registered regardless of what render model they use: /pose/raw and /pose/tip"
@@ -307,8 +307,16 @@ enum BoundPoseType {
     /// "If you provide /pose/tip in your rendermodel you should set it to the position and rotation that are appropriate for pointing (i.e. with a laser pointer) with your controller."
     /// ~https://github.com/ValveSoftware/openvr/wiki/Input-Profiles#pose-components
     Tip,
+    Base,
     /// Not sure why games still use this, but having it be equivalent to raw seems to work fine.
     Gdc2015,
+    Handgrip,
+    Grip,
+    OpenxrHandmodel,
+    OpenxrPinch,
+    OpenxrPoke,
+    OpenxrAim,
+    OpenxrGrip,
 }
 
 macro_rules! get_action_from_handle {
@@ -745,7 +753,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput011_Interface for Input<C> {
                 .map(|h| (hand, h.profile_path))
                 .unzip()
         };
-        let (active_origin, hand) = match loaded.try_get_action(action) {
+        let (active_origin, hand, pose_type) = match loaded.try_get_action(action) {
             Ok(ActionData::Pose) => {
                 let (mut hand, interaction_profile) = match subaction_path {
                     x if x == self.get_subaction_path(Hand::Left) => get_hand(Hand::Left),
@@ -795,7 +803,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput011_Interface for Input<C> {
                     }
                 };
 
-                let Some(ty) = pose_type else {
+                if pose_type.is_none() {
                     trace!("action has no bindings for the hand {hand:?}");
                     no_data!()
                 };
@@ -806,20 +814,13 @@ impl<C: openxr_data::Compositor> vr::IVRInput011_Interface for Input<C> {
                     Hand::Right => self.right_hand_key.data().as_ffi(),
                 });
 
-                match ty {
-                    BoundPoseType::Raw | BoundPoseType::Gdc2015 => (origin, hand),
-                    BoundPoseType::Tip => {
-                        // ToDo: Check if render model has a tip pose otherwise use raw pose
-                        // For now, just use the raw pose
-                        (origin, hand)
-                    }
-                }
+                (origin, hand, pose_type)
             }
             Ok(ActionData::Skeleton(hand)) => {
                 if subaction_path != xr::Path::NULL {
                     return vr::EVRInputError::InvalidDevice;
                 }
-                (0, *hand)
+                (0, *hand, None)
             }
             Ok(_) => return vr::EVRInputError::WrongType,
             Err(e) => return e,
@@ -830,7 +831,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput011_Interface for Input<C> {
 
         unsafe {
             let pose = self
-                .get_controller_pose(hand, Some(origin))
+                .get_controller_pose(hand, Some(origin), pose_type)
                 .unwrap_or_default();
             action_data.write(vr::InputPoseActionData_t {
                 bActive: true,
--- a/src/input/action_manifest/bindings.rs
+++ b/src/input/action_manifest/bindings.rs
@@ -113,7 +113,15 @@ fn parse_pose_binding<'de, D: serde::Deserializer<'de>>(
     let pose = match pose {
         "raw" => BoundPoseType::Raw,
         "tip" => BoundPoseType::Tip,
+        "base" => BoundPoseType::Base,
         "gdc2015" => BoundPoseType::Gdc2015,
+        "handgrip" => BoundPoseType::Handgrip,
+        "grip" => BoundPoseType::Grip,
+        "openxr_handmodel" => BoundPoseType::OpenxrHandmodel,
+        "openxr_pinch" => BoundPoseType::OpenxrPinch,
+        "openxr_poke" => BoundPoseType::OpenxrPoke,
+        "openxr_aim" => BoundPoseType::OpenxrAim,
+        "openxr_grip" => BoundPoseType::OpenxrGrip,
         other => {
             warn!("Unknown pose type: {other:?}");
             BoundPoseType::Raw
--- a/src/input/devices.rs
+++ b/src/input/devices.rs
@@ -214,6 +214,7 @@ impl TrackedDevice {
 
         // Return the raw pose or the transformed pose based on given `pose_type`
         cache_val
+            .filter(|val| val.bPoseIsValid)
             .zip(pose_type)
             .map(|(mut val, p_type)| {
                 val.mDeviceToAbsoluteTracking = vr::HmdMatrix34_t::from(
--- a/src/input/devices.rs
+++ b/src/input/devices.rs
@@ -91,6 +91,7 @@ pub struct TrackedDevice {
     pub profile_path: xr::Path,
     pub connected: bool,
     pub previous_connected: bool,
+    /// Per-frame cache of device's _Raw_ pose.
     pose_cache: Mutex<Option<vr::TrackedDevicePose_t>>,
 }
 
@@ -139,43 +140,10 @@ fn get_controller_pose(
         (xr::SpaceLocation::default(), xr::SpaceVelocity::default())
     };
 
-    // Transform controller location & orientation based on profile's hand+pose transformation
-    // matrix.
-    if let Some(location_transform) = pose_type.and_then(|p_type| {
-        controller.profile_data.as_ref().and_then(|profile_data| {
-            profile_data.pose_transformation(controller.get_controller_hand().unwrap(), p_type)
-        })
-    }) {
-        let raw_pose = location.pose;
-        let pose_mat = Mat4::from_rotation_translation(
-            Quat::from_xyzw(
-                raw_pose.orientation.x,
-                raw_pose.orientation.y,
-                raw_pose.orientation.z,
-                raw_pose.orientation.w,
-            ),
-            Vec3 {
-                x: raw_pose.position.x,
-                y: raw_pose.position.y,
-                z: raw_pose.position.z,
-            },
-        );
-        let (_, new_quat, new_vec) =
-            (pose_mat * location_transform).to_scale_rotation_translation();
-        let [quat_x, quat_y, quat_z, quat_w] = new_quat.to_array();
-        location.pose = xr::Posef {
-            orientation: xr::Quaternionf {
-                x: quat_x,
-                y: quat_y,
-                z: quat_z,
-                w: quat_w,
-            },
-            position: xr::Vector3f {
-                x: new_vec.x,
-                y: new_vec.y,
-                z: new_vec.z,
-            },
-        };
+    // Transform location & orientation based on profile's hand+pose
+    // transformation matrix.
+    if let Some(p_type) = pose_type {
+        location.pose = controller.transform_pose(location.pose, p_type);
     }
 
     Some(vr::space_relation_to_openvr_pose(location, velocity))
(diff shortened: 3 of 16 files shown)


src/input/action_manifest/bindings.rs
```diff
@@ -113,7 +113,15 @@ fn parse_pose_binding<'de, D: serde::Deserializer<'de>>(
...
         "tip" => BoundPoseType::Tip,
+        "base" => BoundPoseType::Base,
         "gdc2015" => BoundPoseType::Gdc2015,
+        "handgrip" => BoundPoseType::Handgrip,
+        "grip" => BoundPoseType::Grip,
+        "openxr_handmodel" => BoundPoseType::OpenxrHandmodel,
+        "openxr_pinch" => BoundPoseType::OpenxrPinch,
+        "openxr_poke" => BoundPoseType::OpenxrPoke,
+        "openxr_aim" => BoundPoseType::OpenxrAim,
+        "openxr_grip" => BoundPoseType::OpenxrGrip,
```
prikhi, line 124: found all these in various `rendermodels` json files. dunno if openvr games would ever use the `openxr_*` poses but figured it doesn't hurt to include them, esp if games support launching in both openvr & openxr.

src/input/profiles/knuckles.rs
```diff
@@ -146,6 +147,159 @@ impl InteractionProfile for Knuckles {
             .inverse(),
         }
     }
+
```
prikhi, line 129: Expand the section above this chunk to see `offset_grip_pose`. It seems like that is just the inverse matrices of `pose_transformation(BoundPoseType::Grip)`? Did I duplicate some existing work? 

Should I kill that function & just use this one? What's the `inverse()` in that call for - is there some math trick that improves performance by using the inverse instead of the actual transformation matrix?

src/input/profiles/knuckles.rs
```diff
@@ -146,6 +147,159 @@ impl InteractionProfile for Knuckles {
...
+    fn pose_transformation(pose: BoundPoseType) -> Option<PoseTransformations> {
+        match pose {
+            BoundPoseType::Raw => None,
+            // All 0 so skip transformation
+            BoundPoseType::Gdc2015 => None,
+            // Not present in rendermodels
+            BoundPoseType::OpenxrAim => None,
+            // Not present in rendermodels
+            BoundPoseType::OpenxrGrip => None,
+            BoundPoseType::Tip => Some(PoseTransformations {
```
prikhi, line 140: I uh, hand-transcribed these from the json so a set of verifying eyes would be nice. I did index->vive->oculus, by the time I got to oculus I was done w/ that so I wrote a script(will paste outside of self-review comments). I could run that script for knuckles & vive if we wanted to be sure I didn't mistype.

src/input/profiles/vive_controller.rs
```diff
@@ -110,6 +110,10 @@ impl InteractionProfile for ViveWands {
     fn offset_grip_pose(_: Hand) -> Mat4 {
         Mat4::IDENTITY
     }
+
+    fn pose_transformation(_pose: BoundPoseType) -> Option<PoseTransformations> {
+        None
+    }
```
prikhi: I _just_ found this in `SteamVR/resources/rendermodels/vr_controller_vive_1_5/vr_controller_vive_1_5.json` I could add it but I dunno why there's one file & not separate left & right matrices... the `openxr_*` ones have `openxr_*_r` matrices but not the normal base/grip/handgrip etc.

src/input/profiles/vive_focus3.rs
```diff
@@ -144,6 +144,139 @@ impl InteractionProfile for ViveFocus3 {
             .inverse(),
         }
     }
+
+    /// Derived from `SteamVR/drivers/vrlink/resources/rendermodels/vive_focus3_controller*/*.json`
+    fn pose_transformation(pose: BoundPoseType) -> Option<PoseTransformations> {
```
prikhi, line 128: This one was also hand-written

src/input/devices.rs
```diff
@@ -50,6 +51,7 @@ impl std::fmt::Debug for TrackedDeviceType {
 pub struct ProfileData {
     properties: &'static ProfileProperties,
     get_hand_offset: fn(Hand) -> Mat4,
+    get_pose_transformation: fn(BoundPoseType) -> Option<PoseTransformations>,
```
prikhi, line 55: I think we need this here since `TrackedDevice` no longer has access to the `InteractionProfile`, just `ProfileData`.

src/input/devices.rs
```diff
@@ -127,6 +139,45 @@ fn get_controller_pose(
...
+                x: quat_x,
+                y: quat_y,
+                z: quat_z,
+                w: quat_w,
+            },
+            position: xr::Vector3f {
+                x: new_vec.x,
+                y: new_vec.y,
+                z: new_vec.z,
+            },
```
prikhi: I dunno if there is a smarter, more efficient way of applying the pose transforms, but I have verified this fixes the issue w/ backwards pointing hands in Dungeons of Eternity.

src/input/devices.rs
```diff
@@ -172,16 +223,19 @@ impl TrackedDevice {
...
         origin: vr::ETrackingUniverseOrigin,
+        pose_type: Option<BoundPoseType>,
     ) -> Option<vr::TrackedDevicePose_t> {
         let mut pose_cache = self.pose_cache.lock().unwrap();
-        if let Some(pose) = *pose_cache {
+        if let Some(pose) = *pose_cache
+            && pose_type.is_none()
+        {
             return Some(pose);
         }
```
prikhi: Multiple interface functions can(& do) call a device's `get_pose` function within a single frame. In Dungeons of Eternity, the `GetPoseActionDataForNextFrame` call happens after calls w/ no `pose_type` defined, which makes the transforms not work since we just pull untransformed poses from the cache. 

So I skipped the fetch if we have a pose_type.

But that opens up some questions:

* Should the cache be indexed by `Option<BoundPoseType>`? What's that look like in rust?
* If we don't cache per-type, should pose_type calls write to the cache or leave it unchanged? 
* Should we require the `pose_type` arg & force the current `None` callers to pass in `Raw`?

src/input.rs
```diff
@@ -806,20 +814,13 @@ impl<C: openxr_data::Compositor> vr::IVRInput011_Interface for Input<C> {
...
 
-                match ty {
-                    BoundPoseType::Raw | BoundPoseType::Gdc2015 => (origin, hand),
-                    BoundPoseType::Tip => {
-                        // ToDo: Check if render model has a tip pose otherwise use raw pose
-                        // For now, just use the raw pose
-                        (origin, hand)
-                    }
-                }
+                (origin, hand, pose_type)
```
prikhi, line 817: I ignored this TODO & just include the pose_type in the return block cause I don't think it's possible to do that transformation in this scope? I think I tried but some required object properties were private.

src/input.rs
```diff
@@ -830,7 +831,7 @@ impl<C: openxr_data::Compositor> vr::IVRInput011_Interface for Input<C> {
 
         unsafe {
             let pose = self
-                .get_controller_pose(hand, Some(origin))
+                .get_controller_pose(hand, Some(origin), pose_type)
```
prikhi, line 834: This is the only place where we actually pass a `pose_type`

prikhi (commented):
I'm not a rust, matrix math, or openvr/openxr expert so lemme know what you need me to fixup here.

prikhi:
This is the script I used to generate the oculus_touch transforms:

```python
#!/usr/bin/env python
"""
./render_models_to_pose_transforms.py <left-rendermodels>.json <right-rendermodels>.json
"""

import json
import sys

def main():
    [left_filename, right_filename] = sys.argv[1:3]
    left_json = {}
    right_json = {}
    with open(left_filename) as left_file:
        left_json = json.load(left_file)['components']
    with open(right_filename) as right_file:
        right_json = json.load(right_file)['components']

    pairs = [
        ("raw", "BoundPoseType::Raw"),
        ("tip", "BoundPoseType::Tip"),
        ("base", "BoundPoseType::Base"),
        ("gdc2015", "BoundPoseType::Gdc2015"),
        ("handgrip", "BoundPoseType::Handgrip"),
        ("grip", "BoundPoseType::Grip"),
        ("openxr_handmodel", "BoundPoseType::OpenxrHandmodel"),
        ("openxr_pinch", "BoundPoseType::OpenxrPinch"),
        ("openxr_poke", "BoundPoseType::OpenxrPoke"),
        ("openxr_aim", "BoundPoseType::OpenxrAim"),
        ("openxr_grip", "BoundPoseType::OpenxrGrip"),
        ]
    for (json_pose, xrizer_pose) in pairs:
        if (json_pose in left_json) != (json_pose in right_json):
            print(f"Found pair in one file but not other: {json_pose}", file=sys.stderr)
            continue
        if not (json_pose in left_json and json_pose in right_json):
            print(f"            {xrizer_pose} => None,")
            continue
        left_pose = left_json[json_pose]['component_local']
        right_pose = right_json[json_pose]['component_local']
        print(f"""
            {xrizer_pose} => Some(PoseTransformations {{
                left_hand: Mat4::from_rotation_translation(
                    Quat::from_euler(
                        EulerRot::XYZ,
                        {left_pose['rotate_xyz'][0]}_f32.to_radians(),
                        {left_pose['rotate_xyz'][1]}_f32.to_radians(),
                        {left_pose['rotate_xyz'][2]}_f32.to_radians(),
                    ),
                    Vec3::new({left_pose['origin'][0]}, {left_pose['origin'][1]}, {left_pose['origin'][2]}),
                ),
                right_hand: Mat4::from_rotation_translation(
                    Quat::from_euler(
                        EulerRot::XYZ,
                        {right_pose['rotate_xyz'][0]}_f32.to_radians(),
                        {right_pose['rotate_xyz'][1]}_f32.to_radians(),
                        {right_pose['rotate_xyz'][2]}_f32.to_radians(),
                    ),
                    Vec3::new({right_pose['origin'][0]}, {right_pose['origin'][1]}, {right_pose['origin'][2]}),
                ),
            }}),
              """)


if __name__ == '__main__':
    main()
```

src/input/profiles/knuckles.rs
prikhi: ran my script against the json & came back w/ identical lines

src/input/profiles/vive_focus3.rs
prikhi: validated this against output from my script

jaoreir:
I tested this patch to fix a controller rotation offset issue with vivecraft. The `tip` transformation worked like a charm for me. I hope this pr gets merged soon :)

src/input/action_manifest/bindings.rs
Supreeeme: I don't really want to include them if they are unused. It makes the code harder to understand imo. I think it'll be easy enough to add them if a game starts using them.

src/input/devices.rs
Supreeeme: It would make the most sense to me to cache the untransformed pose and transform as necessary each time.

src/input/profiles/knuckles.rs
Supreeeme: `offset_grip_pose` is the raw pose transformation. The comment on this function in the `InteractionProfile` trait definition explains why it's the inverse. The transformation matrix presumably transforms from the raw pose to the grip pose, so using the inverse would get us back to the raw pose as expected by games.

Supreeeme (commented):
If you want to put that script in a `resources` folder in the repo or something that would be useful.

src/input/action_manifest/bindings.rs
prikhi: I have ~30 games installed atm & do see some uses, worth keeping around?

```sh
$ rg 'pose/open_xr' **/*.json
Dungeons of Eternity/DoE_Data/StreamingAssets/SteamVR/binding_psvr2.json
151:               "path" : "/user/hand/left/pose/openxr_aim"
155:               "path" : "/user/hand/right/pose/openxr_aim"
159:               "path" : "/user/hand/left/pose/openxr_grip"
163:               "path" : "/user/hand/right/pose/openxr_grip"
# ...
```

src/input/devices.rs
prikhi: makes sense, can do

src/input/profiles/knuckles.rs
prikhi: hmm interesting i'll peek more at it(just saw the definition site similarity, didn't peek at the callsites) - just seemed odd to have & apply the inverse of a transformation that we didn't have or apply 😅 

seems like some are defined as the inverse of `Grip`, some are defined as inverse of `OpenxrGrip`. Tempted to replace their defs with something like `pose_transformation(BoundPoseType::Grip).map_or(Mat4::IDENTITY, ...).inverse()` so that we still have the "this is derived from Grip/OpenxrGrip" annotation but don't have to repeat the actual matrices twice in the code.

prikhi (commented):
thanks for the review, prob won't have time to mess w/ this til the weekend.

prikhi:
Pushed 3 commits to always & only cache Raw poses, simplify the offset_grip_pose implementations, and add the script.

prikhi:
hmm, test is failing because openvr's `HmdMatrix34_t -> Posef -> HmdMatrix34_t` conversion functions are not the identity function :grimacing: 

```rust
#[test]
fn pose_to_from_hmd34_is_identity() {
    let expected = xr::Posef {
        orientation: xr::Quaternionf {
            x: 0.5,
            y: 0.0,
            z: 0.0,
            w: 0.0,
        },
        position: xr::Vector3f {
            x: 0.0,
            y: 0.0,
            z: 0.0,
        },
    };
    assert_eq!(xr::Posef::from(HmdMatrix34_t::from(expected)), expected);
}
```

```sh
---- convert::pose_to_from_hmd34_is_identity stdout ----

thread 'convert::pose_to_from_hmd34_is_identity' (567339) panicked at openvr/src/convert.rs:167:5:
assertion `left == right` failed
  left: Posef { orientation: Quaternionf { x: 0.0, y: 0.0, z: 0.0, w: 0.8660254 }, position: Vector3f { x: 0.0, y: 0.0, z: 0.0 } }
 right: Posef { orientation: Quaternionf { x: 0.5, y: 0.0, z: 0.0, w: 0.0 }, position: Vector3f { x: 0.0, y: 0.0, z: 0.0 } }
```

And I use the assumption of that property here: 

https://github.com/Supreeeme/xrizer/pull/382/changes/1992ecfea7d854ab776fb1f865185c4b23232829#diff-a08e377509cb45c7cab7793e9248a2305c025b918d77ff8d469c5fe6366eabfdR219-R221

prikhi:
ok, i have done some digging. I guess valid rotation quarternions are supposed to have a length of 1 so the fact that conversion isn't reversible for invalid quarternions(`{ 0.5, 0, 0, 0 }` used in the existing test) seems less relevant. 

The thing I shoulda caught is that the position vector in that assertion is all 0, while the test specifies `0.5` for all of em. Some printf debugging showed we're hitting the `bPoseIsValid: false` branch of `space_relation_to_openvr_pose`, so I'm going to tweak this to only transform _valid_ cached poses, and write a test so we exercise both the valid & invalid branches.

prikhi:
Mkay, test should pass now & prob ready for re-review @Supreeeme. Will retest against Dungeons of Eternity tonight.

prikhi:
Updated to account for new MetaTouchPlus profile from #308

src/input/profiles/vive_controller.rs
ImSapphire: The VIVE wands are ambidextrous controllers, there isn't a separate left and right hand controller

prikhi: word, I'll work those in, interesting that openxr ones _do_ have separate inline left/right poses:

```jsonc
{
  // ...
        "openxr_poke": {
            "component_local" : {
                "origin": [-0.038, -0.058, 0.005],
                "rotate_xyz" : [45.00, 0.00, 0.00]
            }
        },
        "openxr_poke_r": {
            "component_local" : {
                "origin": [0.038, -0.058, 0.005],
                "rotate_xyz" : [45.00, 0.00, 0.00]
            }
        },
  // ...
}
```

I haven't seen any games that bind a `openxr_*_r` pose though so I'm guessing it's safe to just put those as the right hand of the `openxr_*` pose instead of adding an R variant of them to the `BoundPoseType`

prikhi: transforms for ViveWand pushed, should we use one of these for the `offset_grip_pose` as well?

prikhi:
Hmm, the only diff between `oculus_touch` changes & `touch_controller_plus`  are the matrices in `pose_transformation`.

What happens if you set the profile's `pose_transformation` to just `fn pose_transformation(pose: BoundPoseType) -> Option<PoseTransformations> { None }`?

PR #386, opened 2026-07-05
system: answer DirectX adapter queries

napanto:
D3D11/D3D12 OpenVR games ask the runtime which GPU the HMD is on before creating their device, and xrizer currently can't answer:

- `GetOutputDevice` only handles `TextureType_Vulkan`. For DirectX texture types it logs "Unsupported texture type" and returns without writing `*pnDevice`. Proton's vrclient forwards the value to the game even in that case, so the game reads uninitialized memory where it expects an adapter LUID.
- `GetDXGIOutputInfo` and `GetD3D9AdapterIndex` are `todo!()`, which aborts the game.
- The `IVRSystem_016` version of `GetOutputDevice` (no `VkInstance` parameter) is also `todo!()`. Proton passes this one through untranslated, so games requesting old interface versions can hit the panic.

This implements all of them from one query: create a short-lived Vulkan instance, get the runtime's physical device through `xrGetVulkanGraphicsDeviceKHR`, and read the LUID from `VkPhysicalDeviceIDProperties` plus the device's position in `vkEnumeratePhysicalDevices` order, which is how DXVK orders its DXGI adapters. Most Linux drivers report `deviceLUIDValid = false`; in that case we write 0, which is what Proton's own vrclient reports when it can't provide a device, and the adapter index falls back to 0, always correct on single GPU machines. The result is cached on `System` so repeated queries don't recreate the instance.

The Vulkan path of `GetOutputDevice` is untouched, except that a null instance (only possible through `IVRSystem_016`) now logs an error and writes 0 instead of panicking.

Tested with Assetto Corsa (D3D11, requests `IVRSystem_014`) under GE-Proton9 with Monado: it previously failed with "D3D11CreateDevice failed" and now runs in VR. Half-Life: Alyx (Vulkan) is unaffected. fakexr gains just enough of the Vulkan API to unit test this; its fake `xrGetVulkanGraphicsDeviceKHR` also now writes its output, which it previously left uninitialized.

### DIFF
--- a/src/system.rs
+++ b/src/system.rs
@@ -1,16 +1,17 @@
 use crate::{
     clientcore::{Injected, Injector},
+    graphics_backends::AdapterInfo,
     input::Input,
     openxr_data::{Hand, RealOpenXrData, SessionData},
     overlay::OverlayMan,
     tracy_span,
 };
 use glam::{Mat3, Quat, Vec3};
-use log::{debug, error, trace, warn};
+use log::{debug, error, info, trace, warn};
 use openvr as vr;
 use openxr as xr;
 use std::ffi::{CStr, CString};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, OnceLock};
 
 #[derive(Copy, Clone)]
 pub struct ViewData {
@@ -156,6 +157,7 @@ pub struct System {
     overlay: Injected<OverlayMan>,
     vtables: Vtables,
     views: Mutex<ViewCache>,
+    adapter_info: OnceLock<Option<AdapterInfo>>,
 }
 
 mod log_tags {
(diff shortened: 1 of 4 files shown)


src/system.rs
```diff
@@ -853,26 +875,70 @@ impl vr::IVRSystem026_Interface for System {
...
+                        .vulkan_graphics_device(self.openxr.system_id, instance as _)
+                        .expect("Failed to get vulkan physical device")
+                        as _;
+                }
+            }
+            // Direct3D games ask for the LUID of the adapter the HMD is
+            // connected to, then create their device on the DXGI adapter with
+            // the matching LUID. Proton usually rewrites this into a Vulkan
+            // query before it reaches us, but the IVRSystem_016 version comes
+            // through untranslated.
```
Supreeeme, line 908: This sounds more like something that needs to be raised with the Proton folks. I don't really want to add stuff that might change at any time within Proton. (if I have to, I will, but I'd rather not.)

Mr-Zero88:
i have made a simple stub fot those in https://github.com/Supreeeme/xrizer/pull/426. as proton does not actually use the output. i will flag this in upstream as well but we need this stub anyways to have older proton work.

PR #388, opened 2026-07-13
Improve overlays, skybox, color conversion and canting

TheIronWolfModding:
Improves overlay, skybox, color, and canting handling.

This is mostly LLM port of stuff I did in OC for racing games.

Improves overlay, skybox, color, and canting handling.

- Overlays: OpenVR games can render overlays without calling Submit or PostPresentHandoff (e.g. menu/HUD-only frames), but xrizer's WaitGetPoses only ended a frame after a Submit. Auto-flush pending overlay-only frames when the state ended at Waited so overlay layers actually reach the compositor. Apply matrix diagonal as scale to quad/cylinder size. Cylinder radius = distance-to-user (wraps around head). Fix cylinder aspect_ratio. Treat SetOverlayCurvature(0) as flat quad. Overlay swapchain arraySize=1.
- Skybox: render forward face as curved panel using pTextures[0], always visible.
- Color: manual sRGB decode in overlay shader to avoid double-encoding through SRGB attachment.
- Canting: opt-in XRIZER_PRESERVE_VIEW_CANTING=1 preserves per-eye canted orientations (needed for Pimax etc.).

### DIFF
--- a/shaders/src/overlay.frag
+++ b/shaders/src/overlay.frag
@@ -3,6 +3,19 @@ layout(set = 0, binding = 0) uniform sampler2D overlay;
 layout(location = 0) in vec2 texCoord;
 layout(location = 0) out vec4 color;
 
+// Manually decode sRGB. The game texture is UNORM but its bytes are
+// sRGB-encoded (typical for game backbuffers). Sampling a UNORM view returns
+// the raw byte value which the hardware treats as linear. We convert here so
+// the sRGB swapchain attachment re-encodes it correctly, giving a lossless
+// round-trip through the compositor.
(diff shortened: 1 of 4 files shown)


galister:
what's the intention behind getting rid of the functional skybox and replacing it with a "curved image in front of the user"? it's called a skybox for a reason.

TheIronWolfModding:
You are right - years ago OC did not have skybox, so what I did there I simply used front quad only (it is used by racing games (rF2 and Raceroom) to display track loading status).  Now, curved simply looks better - allows optionally curved non square texture etc, this is not general, I should not have included it in PR - feel free to abandon.

PR #394, opened 2026-08-02
system: report HMD tracking system, model, and manufacturer strings (Serious Sam 3, seemingly fixes #232 too)

### LINKED ISSUE #232: [Bug]: Serious Sam : The First Encounter VR | Index touchpad not supported
### Operating System

Nobara 42

### OpenXR Runtime

Monado

### Describe the issue.

The game seems to use a legacy input method, which prevents proper full usage of the Index controllers. As a matter of fact, it still displays HTC Vive controllers.
Interestingly, the game defaults to a teleport mode by clicking the joystick, which works well. But when switching to a “Classic” mode, which should use those same joysticks to freely move around, it doesn't work and seems to require the touchpad for those. I would ideally rebind those, but they don't allow you to rebind those controls in-game for the Index Controller itself.

### Steps to reproduce

1. Launch [Serious Sam : The First Encounter VR](https://store.steampowered.com/app/552450/Serious_Sam_VR_The_First_Encounter/) (If you don't have it, I believe you can try to use the [Serious Sam Fusion](https://store.steampowered.com/app/564310/Serious_Sam_Fusion_2017_beta/) cross-game platform to load the demo level in VR)
2. Launch the game
3. Try moving (Teleport using cliquing the joystick)
4. Change the controls to "Classic".
5. Try moving again.

skryvel:
Game in question: Serious Sam 3 VR: BFE via Steam and Proton CachyOS from July
Hardware: Quest 3 on KDE Wayland via WiVRN and opensource AMD drivers
Prior status: shows some Vive looking controllers, asks to press right trigger, nothing happens with any button
Current status: fully playable, both controllers, shows Oculus models
Maybe odd: some in-gameplay interface objects have a strong glow - maybe that's just how it is
Note: game complains about display driver, but allows it to be ignored and works fine
Code by: Fable 5

---

Some games read the HMD's TrackingSystemName/ModelNumber/ManufacturerName strings to decide which motion controller scheme to enable. When these come back empty, such games conclude no supported controllers are present and ignore all controller input (menus still work via mouse) while rendering their fallback wand models.

Report Oculus-style identity strings for the HMD, consistent with the tracking system name already reported for the Touch controller profiles.

### DIFF
--- a/fakexr/src/lib.rs
+++ b/fakexr/src/lib.rs
@@ -289,7 +289,7 @@ pub unsafe extern "system" fn get_instance_proc_addr(
                     (ResultToString),
                     (StructureTypeToString),
                     (GetInstanceProperties),
-                    (GetSystemProperties),
+                    GetSystemProperties,
                     CreateSwapchain,
                     DestroySwapchain,
                     EnumerateSwapchainImages,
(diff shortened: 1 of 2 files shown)


skryvel:
> Questioned in another PR, the hardcoding of `oculus` etc.

OpenComposite reference. Also apparently these games read the HMD identity strings once at startup and never re-query. In my traces the game asked for TrackingSystemName ~14 seconds before the runtime bound any controller interaction profile, so at answer time there's no profile to pull from. OpenComposite probably had this too and hardcodes "oculus"/"Oculus". Games that identify controllers still get the profile-accurate strings from the controller devices, this fixed string only feeds the HMD-sniffing path.

ImSapphire:
I don't think it's a good idea to hard-code the Quest 3 properties here. One idea I had is to spin up a headless session in IVRClientCore::Init and create a dummy action set with a single binding, suggest bindings for all interaction profiles, call xrSyncActions, then cache the bound interaction profiles and destroy the headless session before continuing initialisation. Then we'd have the information available early when games request such properties immediately.

PR #395, opened 2026-08-03
hl2-vr: multiple fixes (fixes #266) (fixes #369)

### LINKED ISSUE #266: [Bug]: Half Life 2 VR Mod loads, but some actions are unbound
### Operating System

EndevourOS (newest version, Gnome on Wayland)

### OpenXR Runtime

WiVrn

### Describe the issue.

after starting a game in HL2 VR Mod i noticed that I can't grab items. firstly I thought that maybe it's a normal thing (since I was just at the start, leaving the train), but i had "show tutorial hints" on and after I got to the place where you meet barney the tutorial hints told me that grabbing objects was unbound for some reason

i tried manually bind the grip button on the oculus touch controllers (I have an Oculus Quest 2) to grabbing in the xrizer folder in the oculus_touch.json file but it didn't help

here's the log if you find it helpful:

[xrizer.txt](https://github.com/user-attachments/files/24389596/xrizer.txt)

### Steps to reproduce

install 32bit compatible versions of WiVrn and xrizer, install HL2 VR Mod, connect Oculus Quest 2 to the computer using WiVrn and play HL2 VR Mod with these launch options:  "PRESSURE_VESSEL_IMPORT_OPENXR_1_RUNTIMES=1 %command% "

### LINKED ISSUE #369: [Bug]: Half-Life 2 VR Mod Crashes when Antialiasing is disabled
### Operating System

Arch Linux

### OpenXR Runtime

WiVRn

### Describe the issue.

When MSAA is disabled, either at startup (from an existing `config.cfg`) or attempting to change the MSAA setting in-game, xrizer will crash from "invalid usage" of `SwapchainCreateFlags(0)`.

[xrizer log from a working game start, triggering crash from disabling MSAA](https://github.com/user-attachments/files/28675516/xrizer.txt)

###### Log is technically from #359, but its current state is up-to-date with upstream as of writing.

The same exact crash occurs with the Episodic releases as well, using the base game version as the example.

### Steps to reproduce

0. Ensure `~/.steam/steam/userdata/{USERID}/658920/remote/cfg/config.cfg` is either deleted, or the string `hlvr_msaa` is not set to `"1"` (corresponds to "Antialiasing mode: None"
1. Start HL2VR with xrizer
2. Navigate to *Options->Performance tab->Antialiasing mode*, which usually defaults to *4x MSAA", and set to *None*
3. Click "Apply"; observe the xrizer crash

Crash does not happen when setting to any other Antialiasing mode, but only when set to *None* specifically.

skryvel:
Status: secomd iteration but I haven't retested and reviewed; dropped the tip commit

Played Half-Life 2: VR Mod (32-bit) end to end and fixed everything hit along the way. Each commit stands alone. Tested on Quest 3 via WiVRn, under both `oculus/touch_controller` and `meta/touch_controller_plus` (WiVRn picked different profiles across sessions).  KDE/Wayland/Proton/D3D9. PR is a bit big, happy to split or whatever, just let me know.

Test session time: 15m + troubleshooting iterations (early chapter 2).

### Grabbing did not work (#266)

HL2VR's `bindings_touch.json` binds grip with mode `"button"` and a nonstandard `"grab"` input slot instead of `"click"`, which was silently dropped during deserialization.

**Before:** grabbing unbound, the in-game tutorial hints say as much.
**After:** the `grab` slot is translated like `click` (squeeze value + click thresholds on force-less controllers). Picking up and throwing objects works.

### Binding hints showed "UNBOUND" for everything

HL2VR resolves its `%+command%` hint tokens through `GetActionOrigins` + `GetOriginTrackedDeviceInfo`, both stubs. Games read the hand from `devicePath` (compared against `GetInputSourceHandle` results) and the button from `rchRenderModelComponentName`.

**Before:** every hint reads "UNBOUND ..." (or renders uninitialized buffer contents).
**After:** `GetActionOrigins`, `GetOriginLocalizedName` and `GetOriginTrackedDeviceInfo` are implemented, backed by `xrEnumerateBoundSourcesForAction` / `xrGetInputSourceLocalizedName`, including the internal threshold/toggle/grab actions that custom bindings resolve to. Component names come from the SteamVR render model jsons matching each profile's `render_model_name`. Hints show the right buttons.

### Laser pointer / menu cursor pointed ~40deg off

`/pose/tip` was a TODO returning the raw pose.

**Before:** the cursor is only usable by tilting the controller ~90deg (or at all, used left controller instead of right, difficult to make it respond).
**After:** each hand gets a tip space using the `tip` component transform from the controller's SteamVR render model json (Touch / Touch Plus). Profiles without a known tip transform keep the raw pose. Cursor moves with right  controller just fine.

### Crash with MSAA disabled (#369)

Games submit `m_nSampleCount = 0` when MSAA is off; OpenXR requires at least 1, so swapchain creation failed and xrizer panicked.

**Before:** instant crash with antialiasing set to None.
**After:** sample count is clamped to 1, also before swapchain reuse comparisons so a 0 can't force per-frame recreation.

### Smaller fixes

- Concurrent xrizer processes (a launcher's VR probe alongside the game) truncated the same log file, destroying each other's output. Logs are now per-process (`xrizer-<pid>.txt`) and the crash dialog opens the right file.
- The crash dialog's "Copy to clipboard" usually pasted empty: miniquad's clipboard only lives while the window is open (and is unimplemented on some backends). It now prefers wl-copy/xclip/xsel and falls back to miniquad.
- The dynamic loader panic in a 32-bit build now says it needs a 32-bit OpenXR loader (or `--features static-openxr`) instead of the generic message.
- fakexr implements `xrEnumerateBoundSourcesForAction` / `xrGetInputSourceLocalizedName` so the origin APIs are testable. New tests cover the grab slot, origin queries, tip pose, and zero sample count.

<details>

<summary>Screenshots</summary>

AA off is now fine.
image

Bound keys.
image

Grabbing with right hand.
image

</details>

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -169,7 +169,8 @@ impl Compositor {
         {
             let b_texture =
                 G::get_texture(texture).ok_or(vr::EVRCompositorError::InvalidTexture)?;
-            let info = backend.swapchain_info_for_texture(b_texture, bounds, texture.eColorSpace);
+            let info =
+                backend.checked_swapchain_info_for_texture(b_texture, bounds, texture.eColorSpace);
             Ok(TempBackendData {
                 backend,
                 swapchain_create_info: Some(info),
@@ -1230,9 +1231,9 @@ impl<G: GraphicsBackend> FrameController<G> {
 
         self.eyes_submitted[eye as usize] = if self.should_render {
             // Make sure our image dimensions haven't changed.
-            let new_info = self
-                .backend
-                .swapchain_info_for_texture(texture, bounds, color_space);
+            let new_info =
+                self.backend
+                    .checked_swapchain_info_for_texture(texture, bounds, color_space);
 
             is_valid_swapchain_info(&new_info)
                 .then(|| {
@@ -1417,6 +1418,7 @@ mod tests {
         static SWAPCHAIN_WIDTH: Cell<u32> = const { Cell::new(10) };
         static SWAPCHAIN_HEIGHT: Cell<u32> = const { Cell::new(10) };
         static SWAPCHAIN_FORMAT: Cell<u32> = const { Cell::new(0) };
+        static SWAPCHAIN_SAMPLE_COUNT: Cell<u32> = const { Cell::new(1) };
     }
 
     pub enum FakeApi {}
@@ -1482,7 +1484,7 @@ mod tests {
                 create_flags: xr::SwapchainCreateFlags::EMPTY,
                 usage_flags: xr::SwapchainUsageFlags::EMPTY,
                 format: SWAPCHAIN_FORMAT.get(),
-                sample_count: 1,
+                sample_count: SWAPCHAIN_SAMPLE_COUNT.get(),
                 width: SWAPCHAIN_WIDTH.get(),
                 height: SWAPCHAIN_HEIGHT.get(),
                 face_count: 1,
@@ -1753,6 +1755,31 @@ mod tests {
             (&raw mut (*timing.as_mut_ptr()).m_nSize).write(0);
         }
         assert!(!f.comp.GetFrameTiming(timing.as_mut_ptr(), 1));
+    }
+
+    #[test]
+    fn zero_sample_count_texture() {
+        // Games submit sample count 0 when MSAA is disabled (e.g. HL2VR with
+        // antialiasing set to None) - it should be treated as 1, not crash
+        // swapchain creation.
+        let f = Fixture::new();
+        SWAPCHAIN_SAMPLE_COUNT.set(0);
+
+        assert_eq!(f.wait_get_poses(), None);
+        assert_eq!(f.submit(vr::EVREye::Left), None);
+        assert_eq!(f.submit(vr::EVREye::Right), None);
+
+        assert_eq!(f.wait_get_poses(), None);
+        {
+            let data = f.comp.openxr.session_data.get();
+            let lock = data.comp_data.0.lock().unwrap();
+            let DynFrameController::Fake(ctrl) = lock.as_ref().unwrap() else {
+                panic!("Frame controller was not set up or not faked!");
+            };
+            let swapchain_data = ctrl.swapchain_data.as_ref().expect("no swapchain created");
+            assert_eq!(swapchain_data.info.sample_count, 1);
+        }
+        SWAPCHAIN_SAMPLE_COUNT.set(1);
     }
 
     #[test]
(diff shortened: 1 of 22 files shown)


src/compositor.rs
```diff
@@ -169,7 +169,10 @@ impl Compositor {
         {
             let b_texture =
                 G::get_texture(texture).ok_or(vr::EVRCompositorError::InvalidTexture)?;
-            let info = backend.swapchain_info_for_texture(b_texture, bounds, texture.eColorSpace);
+            let mut info =
+                backend.swapchain_info_for_texture(b_texture, bounds, texture.eColorSpace);
+            // 0 means MSAA off - OpenXR requires at least 1
+            info.sample_count = info.sample_count.max(1);
```
Mr-Zero88: this is duplicated as swapchain_info_for_texture already does this check

src/compositor.rs
```diff
@@ -1230,9 +1233,11 @@ impl<G: GraphicsBackend> FrameController<G> {
 
         self.eyes_submitted[eye as usize] = if self.should_render {
             // Make sure our image dimensions haven't changed.
-            let new_info = self
+            let mut new_info = self
                 .backend
                 .swapchain_info_for_texture(texture, bounds, color_space);
+            // 0 means MSAA off - OpenXR requires at least 1
+            new_info.sample_count = new_info.sample_count.max(1);
```
Mr-Zero88: same here as swapchain_info_for_texture will clamp to 1 already. can drop the mut as well

Mr-Zero88 (changes requested):

Mr-Zero88 (commented):
this needs a cleanup to not spam folder with logs

src/input/profiles.rs
```diff
@@ -561,9 +569,13 @@ pub mod paths {
                 "b" => Some(Self::B),
                 "x" => Some(Self::X),
                 "y" => Some(Self::Y),
-                "application_menu" => Some(Self::Menu),
+                // "menu"/"squeeze" and friends are the OpenXR spellings; accepting both
```
Mr-Zero88: since this funciton is named from_openvr_str this is not realy acutrate. could you split this into a from_openvr_str and a from_openxr_str?

Mr-Zero88:
i think it would be good to split this into multiple PRs.

- clipboard
- 32bit openxr loader
- swapchain sample count
- per process logging
- pose offsets (this needs further discussion)
- action origins and fakexr implementation combined
- grab button binding support

Mr-Zero88:
also make shure to fix up all the commits with `cargo fmt` to have the correct format

src/lib.rs
```diff
@@ -142,7 +199,14 @@ fn init_logging() {
                 };
 
                 match std::fs::create_dir_all(&path) {
-                    Ok(_) => setup(),
+                    Ok(_) => {
+                        prune_old_logs(
+                            &path,
+                            std::time::SystemTime::now()
+                                - std::time::Duration::from_secs(48 * 60 * 60),
```
ImSapphire:
```suggestion
                                - std::time::Duration::from_hours(48),
```

prikhi:
Tip pose transformation should be fixed by #382

PR #397, opened 2026-08-06
hl2-vr: multiple fixes

skryvel:
Continuation of #395, branch was deleted by accident; all review feedback from there is hopefully addressed.

Status: good for review
New:
 - fourth iteration: no new throwing improvement, seems as good as it gets with the workaround; the PR has been split, not sure if there's more worth splitting
 - third iteration: two pieces split off into new pr's, feedback addressed, throwing improved, commit from Serious Sam 3 PR removed
 - second iteration: some work dropped in favor of (better) work from another live PR, throwing is now also fixed(ish), and a commit from a Superhot VR fix (not PR'ed yet) that helped something here that I forgot. ~Rust version also bumped because `from_hours`.~ (deferred)

Played Half-Life 2: VR Mod (32-bit) end to end and fixed everything hit along the way. Each commit stands alone. Tested on Quest 3 via WiVRn, under both `oculus/touch_controller` and `meta/touch_controller_plus` (WiVRn picked different profiles across sessions).  KDE/Wayland/Proton/D3D9. PR is a bit big, happy to split or whatever, just let me know.

Test session time: 15m + troubleshooting iterations (early chapter 2).

### Binding hints showed "UNBOUND" for everything

HL2VR resolves its `%+command%` hint tokens through `GetActionOrigins` + `GetOriginTrackedDeviceInfo`, both stubs. Games read the hand from `devicePath` (compared against `GetInputSourceHandle` results) and the button from `rchRenderModelComponentName`.

**Before:** every hint reads "UNBOUND ..." (or renders uninitialized buffer contents).
**After:** `GetActionOrigins`, `GetOriginLocalizedName` and `GetOriginTrackedDeviceInfo` are implemented, backed by `xrEnumerateBoundSourcesForAction` / `xrGetInputSourceLocalizedName`, including the internal threshold/toggle/grab actions that custom bindings resolve to. Component names come from the SteamVR render model jsons matching each profile's `render_model_name`. Hints show the right buttons.

### Throwing

Noticed throwing was a bit off and eventually got it workable by simulating the velocity vector from positions. Not perfect but also with HL2 janky physics I can't tell. It got pretty decent in the end.

### Smaller fixes

- fakexr implements `xrEnumerateBoundSourcesForAction` / `xrGetInputSourceLocalizedName` so the origin APIs are testable. New tests cover the grab slot, origin queries, tip pose, and zero sample count.

<details>

<summary>Screenshots</summary>

Bound keys.
image

</details>

<details>

[...]

PR #398, opened 2026-08-08
compositor: treat swapchain sample count 0 as 1 (fixes #369)

### LINKED ISSUE #369: [Bug]: Half-Life 2 VR Mod Crashes when Antialiasing is disabled

skryvel:
Split out of #397 as its own change.

Games submit a sample count of 0 when MSAA is disabled — Half-Life 2: VR does it whenever antialiasing is set to None, either from the config file or the in-game setting. OpenXR requires at least 1, so the runtime rejects the swapchain and xrizer panics before the game ever renders.

**Before:** crash on startup with antialiasing set to None.
**After:** the game runs.

The clamp lives in a `checked_swapchain_info_for_texture` wrapper on the backend trait, so there is a single place doing it and the fake backend used by the tests goes through it too. It is applied before the create info is compared against the current swapchain as well — clamping only at creation would leave a stored 1 mismatching the raw 0 arriving every frame, quietly recreating the swapchain on every submit.

Covered by `zero_sample_count_texture`. Tested with Half-Life 2: VR on a Quest 3 over WiVRn.

image

### DIFF
--- a/src/graphics_backends.rs
+++ b/src/graphics_backends.rs
@@ -19,12 +19,28 @@ pub trait GraphicsBackend: Into<SupportedBackend> {
     /// Returns None if the texture is invalid.
     fn get_texture(texture: &vr::Texture_t) -> Option<Self::OpenVrTexture>;
 
-    fn swapchain_info_for_texture(
+    /// Backend implementation for swapchain_info_for_texture. Callers should
+    /// use swapchain_info_for_texture instead, which fixes up invalid values
+    /// games are known to submit.
+    fn raw_swapchain_info_for_texture(
         &self,
         texture: Self::OpenVrTexture,
         bounds: vr::VRTextureBounds_t,
         color_space: vr::EColorSpace,
     ) -> xr::SwapchainCreateInfo<Self::Api>;
+
+    fn swapchain_info_for_texture(
+        &self,
+        texture: Self::OpenVrTexture,
+        bounds: vr::VRTextureBounds_t,
+        color_space: vr::EColorSpace,
+    ) -> xr::SwapchainCreateInfo<Self::Api> {
+        let mut info = self.raw_swapchain_info_for_texture(texture, bounds, color_space);
+        // Games submit a sample count of 0 when MSAA is disabled (e.g. Half-Life
+        // 2: VR with antialiasing set to None), but OpenXR requires at least 1.
+        info.sample_count = info.sample_count.max(1);
+        info
+    }
 
     fn store_swapchain_images(
         &mut self,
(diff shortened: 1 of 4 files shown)


src/graphics_backends.rs
```diff
@@ -26,6 +26,22 @@ pub trait GraphicsBackend: Into<SupportedBackend> {
         color_space: vr::EColorSpace,
     ) -> xr::SwapchainCreateInfo<Self::Api>;
 
+    /// Like swapchain_info_for_texture, but with invalid values games are known
+    /// to submit fixed up. Callers creating or comparing swapchains should use
+    /// this to keep the create info consistent across frames.
+    fn checked_swapchain_info_for_texture(
```
ImSapphire: Why is this a separate function? The only two users of `swapchain_info_for_texture` are changed to call the checked version in this PR, you could just do the clamp inside `swapchain_info_for_texture`

ImSapphire: Ah, nevermind I see why. Though I think having a 'checked' version is weird, I think `swapchain_info_for_texture` itself should be a wrapper around a trait function, since it's possible for someone to mistakenly call `swapchain_info_for_texture` and cause problems

skryvel: Indeed. Reworked. Backends now implement raw_swapchain_info_for_texture, and swapchain_info_for_texture is the trait wrapper that does the fixup so there's no unchecked version left to call by mistake. Turns out that overlay.rs was already making that exact mistake.

PR #406, opened 2026-08-14
input: action set name handling invalid characters, allow string or number in json binding [Halo CEVR]

Mr-Zero88:


src/input/action_manifest/bindings.rs
```diff
@@ -239,33 +239,73 @@ where
...
-        let ret = <&str>::deserialize(deserializer)?;
-        ret.parse().map_err(|_| {
-            D::Error::custom(format_args!(
-                "invalid value: expected {}, got {ret}",
-                std::any::type_name::<T>()
-            ))
-        })
+        use serde::de::{self, Visitor};
+
+        struct StringOrTypedVisitor<T>(PhantomData<T>);
```
Supreeeme, line 267: I hate this visitor thing and I think we can avoid doing it, if we just have `StringOrTyped` be an enum:
```rust
#[derive(Deserialize)]
#[serde(untagged)]
enum StringOrTyped<T> {
    String(String)
    Typed(T)
}
```
and have a method like `fn value(&self) -> T`

PR #412, opened 2026-08-18
PAYDAY 2: Fixed GetOverlayTransformAbsolute

skulleven:
Booting PAYDAY 2 crashes XRizer with a panic at `src/overlay.rs:1102:9: not yet implemented`.

I was able to bypass it. However the game still crashes xrizer when you want to open up a chat window in game with a panic at 'src/overlay.rs:799:9: not yet implemented' . 

The game works with no issues aside from that.

### DIFF
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -1099,7 +1099,8 @@ impl vr::IVROverlay028_Interface for OverlayMan {
         _: *mut vr::ETrackingUniverseOrigin,
         _: *mut vr::HmdMatrix34_t,
     ) -> vr::EVROverlayError {
-        todo!()
+        crate::warn_unimplemented!("GetOverlayTransformAbsolute");
+        vr::EVROverlayError::None
     }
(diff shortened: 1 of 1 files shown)


src/overlay.rs
```diff
@@ -1099,7 +1099,8 @@ impl vr::IVROverlay028_Interface for OverlayMan {
         _: *mut vr::ETrackingUniverseOrigin,
         _: *mut vr::HmdMatrix34_t,
     ) -> vr::EVROverlayError {
-        todo!()
+        crate::warn_unimplemented!("GetOverlayTransformAbsolute");
```
ImSapphire, line 1102: If we're not going to actually implement it, we should at least put sane values in the output universe and matrix. We do have the transform stored in overlays though, we can grab it from there.

Supreeeme:
Sorry for the delay. Thanks for the contribution, but I'm going to accept #417 over this one.

Issue #281, opened 2026-01-16
[Bug]: Can't throw objects in Half-Life Alyx


tuffnerdstuff:
### Operating System

EndeavorOS (Arch Linux)

### OpenXR Runtime

WiVRn

### Describe the issue.

OS: EndeavorOS (Basically Arch Linux)
Headset: Meta Quest 2
xrizer: 0.4

As ALVR recently stopped working for me, I looked around for alternatives. I stumbled accross WiVRn and got it to work both from flatpak and from AUR (setup is a little more involved but works). I'm usign it to run Half-Life Alyx and was surprised how well it runs, both wireless and wired. 

I'm having one issue though, which makes progressing really hard: I can't throw objects. When I grab an object (e.g. a bottle), swing my arm and then release the grip, it just falls to the ground. It looks like the motion vector is completely discarded on release.

I did some tests to narrow down the possible cause:
- Throwing worked in Alyx before using ALVR
- With WiVRn + xrizer it does not work, no matter the config
  - Proton Experimental/Hotfix/10/9/8 wireless/wired
  - Native Linux wireless/wired
- Throwing objects works with WiVRn + xrizer in The Lab, though

I also did the usual AI "vibe googling", but it mostly hallucinated configs that were not applicable to xrizer. It mentioned though, that the cause could be that games like "The Lab" feed the physics simulation with raw motion data from the controllers (i.e. velocity is calculated from motion), while Alyx might rely on velocity data from skeletal pose which might not be set by xrizer. Curios, I started looking into closed issues from ALVR and found this [closed issue](https://github.com/alvr-org/ALVR/pull/2416), which addresses a similar issue. Not sure if it is useful, but I wanted to provide anything I could found.

Logs:
[xrizer.txt](https://github.com/user-attachments/files/24681427/xrizer.txt)
Don't mind the panic at the end, I selected controller settings which forwards to SteamVR, which xrizer cannot provide of course.

### Steps to reproduce

- Start Half-Life Alyx
- Grab any object
- Try to throw it (in an arc, by pushing, doesn't matter)
- When releasing the grip the object just falls straight down to the ground

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

tuffnerdstuff:
[...]

PR #416, opened 2026-08-31
input: report the release edge of custom bindings

schlegp:
Was investigating why throwing items in Half-Life: Alyx only worked with the grip grab and not the trigger/pinch grab.

Alyx binds its grab action (/actions/interact/in/use) to two inputs at once. In bindings_touch.json the grip is a trigger mode click, which xrizer binds natively, and the trigger is a button mode click, which becomes a threshold binding on trigger/value.

GetDigitalActionData only adopted the custom binding's state while it was held. Releasing the trigger failed that condition, so it fell back to reporting the natively bound grip instead. The action correctly read as no longer held, but bChanged was false and last_change_time came from whenever the grip last moved — tracing it, the timestamp reported on a trigger release was literally the previous grip release's. Alyx appears to act on that release edge, so it never threw the object, it just let go of it. Grip grabs worked because there the native binding was the one that changed.

Both candidates are now ranked: a pressed binding wins, and between two unpressed ones the one that changed during this sync wins. Applied the same rule in the two places that pick among an action's custom bindings and between hands, which previously fell back to whichever candidate came first.

Includes a regression test using the same binding shape (one native click, one threshold binding). Against the old code it fails with bChanged: false.

Fixes throwing in Alyx (Issue #281). Haven't checked other games as I don't have other similar ones to play.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -164,27 +164,16 @@ impl<C: openxr_data::Compositor> Input<C> {
         debug_assert!(self.left_hand_key.0.as_ffi() != 0);
         debug_assert!(self.right_hand_key.0.as_ffi() != 0);
         let left_state = self.state_from_bindings(action, self.left_hand_key.0.as_ffi());
-
-        match left_state {
-            None => self.state_from_bindings(action, self.right_hand_key.0.as_ffi()),
-            Some((left, _)) => {
-                if left.is_active && left.current_state {
-                    return left_state;
-                }
-                let right_state = self.state_from_bindings(action, self.right_hand_key.0.as_ffi());
-                match right_state {
-                    None => left_state,
-                    Some((right, _)) => {
-                        if right.is_active && right.current_state {
-                            return right_state;
-                        }
-                        if left.is_active {
-                            return left_state;
-                        }
-                        right_state
-                    }
+        let right_state = self.state_from_bindings(action, self.right_hand_key.0.as_ffi());
+        match (left_state, right_state) {
+            (Some((left, _)), Some((right, _))) => {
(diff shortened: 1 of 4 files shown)


src/input/tests.rs
```diff
@@ -1085,3 +1085,67 @@ fn load_actions_race() {
...
+
+/// An action bound to a native input *and* a custom (threshold) binding must report the release
+/// edge of whichever one was actually being used.
+///
+/// Half-Life: Alyx binds its grab action this way - grip as a `trigger` mode click, which becomes a
+/// native binding, and trigger as a `button` mode click, which becomes a threshold on
+/// trigger/value. Releasing the threshold binding used to fall back to reporting the native
+/// binding's state: the action correctly read as no longer held, but `bChanged` was false and
+/// `fUpdateTime`'s companion `last_change_time` came from whenever the *grip* last moved. Alyx
+/// never saw the release edge, so held objects dropped straight down instead of being thrown.
```
Supreeeme, line 1097: This can go in the commit message instead of this comment.

src/input.rs
```diff
@@ -218,22 +207,32 @@ impl<C: openxr_data::Compositor> Input<C> {
...
+                break;
             }
         }
 
         best_state.map(|x| (x, restrict_to_device))
     }
 }
 
+/// Ranking used to decide which of an action's bindings should represent it. Highest wins.
+fn binding_rank(state: &xr::ActionState<bool>) -> (bool, bool, i64) {
```
Supreeeme, line 228: I think it would be simpler if this just took two action states and returned the "highest" one, which should probably just be which of the last_change_times is the latest.

PR #417, opened 2026-09-01
implement GetOverlayTransformAbsolute

galister:
needed for PayDay 2 to not crash

### DIFF
--- a/src/overlay.rs
+++ b/src/overlay.rs
@@ -1095,11 +1095,26 @@ impl vr::IVROverlay028_Interface for OverlayMan {
     }
     fn GetOverlayTransformAbsolute(
         &self,
-        _: vr::VROverlayHandle_t,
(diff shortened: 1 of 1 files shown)


src/overlay.rs
```diff
@@ -1095,11 +1095,35 @@ impl vr::IVROverlay028_Interface for OverlayMan {
...
+            // if none, report the identity in the current tracking space
+            (
+                self.openxr.get_tracking_space(),
+                vr::HmdMatrix34_t {
+                    m: [
+                        [1.0, 0.0, 0.0, 0.0],
+                        [0.0, 1.0, 0.0, 0.0],
+                        [0.0, 0.0, 1.0, 0.0],
+                    ],
+                },
```
Supreeeme:
```suggestion
    xr::Posef::IDENTITY.into()
```

galister:
applied, thanks.

PR #418, opened 2026-09-01
input: implement GetActionOrigins, GetActionBindingInfo and InputProfilePath

schlegp:
No Man's Sky did not properly draw the input glyphs in menus.

Added tracing throughout the input chain to find these functions responsible for input glyphs.

Really don't have experience with Rust, so might be terrible code, but it works at least.

### DIFF
--- a/fakexr/src/lib.rs
+++ b/fakexr/src/lib.rs
@@ -113,6 +113,19 @@ fn get_hand_data(hand: UserPath, session: &Session) -> &HandData {
     }
 }
 
+/// Make xrEnumerateBoundSourcesForAction answer with the bindings suggested for the profile in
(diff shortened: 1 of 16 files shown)


rWolstenholme:
Nice! I can't code in rust, but I was vibe-troubleshooting and this solved my problem.

This fix resolves unresponsive controller inputs (trigger clicks and face buttons not interacting with UI) in SKYBOX VR Video Player (AppID `721090`).
SKYBOX relies on the HTC Vive Input Utility (VIU) (`SteamVRv2Module.cs`), which requires `IVRInput::GetActionOrigins()` and `GetActionBindingInfo()`. Building with this PR applied fully restored trigger clicking and button functionality on a Meta Quest 2 streaming via WiVRn on Linux.

PR #419, opened 2026-09-01
System button

Mr-Zero88:
add system button as a valid path and add it to the correct allowed paths


Supreeeme:
System bindings are not guaranteed to be usable by the application, according to the spec.

Mr-Zero88:
correct but we should let the runtime decide that and still translate the bindings.

Supreeeme:
The problem is that "not usable by the application" means that `xrSuggestInteractionProfileBindings` will actually fail. If you want to add some plumbing to retry without the system bindings whenever it fails I would accept that.

Mr-Zero88:
How I read it is that it has to accept the binding but just ignore it.

I tried it with Monado and with games that suggest it, and it did not fail in xrSuggestInteractionProfileBindings.

PR #422, opened 2026-09-07
input: support Vector2 binding inversion

juriyivanov:
## Summary

Add support for per-source Vector2 axis inversion from SteamVR/OpenVR binding
parameters:

```json
"parameters": {
  "invert": "y"
}
```
Supported values are `x`, `y`, and `xy` for both joystick and trackpad
position bindings.

Bindings without inversion continue to use the existing direct OpenXR
suggested-binding path. Transformed sources use independent backing actions,
so inversion remains isolated by source, interaction profile, and subaction
path, and transformed and direct sources can coexist on the same OpenVR
action.

Vector2 snapshots and deltas are updated from the transformed aggregate during
`UpdateActionState`, preserving delta semantics across repeated reads and
unread frames.

## Motivation

xrizer currently ignores the Vector2 `invert` parameter in SteamVR/OpenVR
binding JSON. Some application bindings rely on this transform to map physical
controller axes to the coordinates expected by the application.

Related to #149.

## Validation

* `cargo fmt --check`
* `git diff --check`
* 12 focused Vector2 inversion tests
* `cargo test` — 95 unit tests + smoke test
* `cargo +1.97 clippy --workspace --all-targets`
* `cargo xbuild --release --features static-openxr --verbose`

Runtime hardware test:

* PICO 4
* WiVRn OpenXR runtime
* The Walking Dead: Saints & Sinners
* Oculus Touch binding using `"invert": "y"`

Before this change the locomotion Y axis was reversed.
With the patch, forward/backward locomotion is correct on hardware.


Supreeeme:
There are entirely too many test cases here. If the inversion code is doing the same thing for each input, it would be enough to make sure every possible input gets parsed properly, and then a test to verify invert actually works.

PR #426, opened 2026-09-15
fix new HalfLife alyx release

Mr-Zero88:
Stubs the GetDXGIOutputInfo and GetD3D9AdapterIndex to return 0.
Proton calls GetD3D9AdapterIndex but ignores its return value so what we return is irelevant. For GetDXGIOutputInfo it does actualy pass though the output but a users has tested it and returning 0 there as well also solves it.

This also adds a interface forward form 16 to 22 for GetOutputDevice.

Updates openxrs to latest git to include https://github.com/Ralith/openxrs/issues/224 fix.


imbusy:
:+1: it worked yesterday at noon Pacific time and then Alyx broke in the evening.

imbusy:
Tested the change and it works! Then tested against the latest release and it crashed with not implemented error. So seems to have a positive effect.

Cargo.toml
```diff
@@ -80,3 +80,7 @@ libloading = "0.8.5"
 test-cdylib = "1.1.0"
 fakexr = { path = "fakexr" }
 macros = { path = "macros", features = ["test"]}
+
+[patch.crates-io]
+openxr = { git = "https://github.com/Ralith/openxrs", package = "openxr" }
+openxr-sys = { git = "https://github.com/Ralith/openxrs", package = "openxr-sys" }
```
Supreeeme: Looks like there was a new release so we can just use that.

Supreeeme (commented):
Looks like the tests need to be updated?

Cargo.toml
Mr-Zero88: Yes, I will use that version and change it right now. Maybe we will have a conflict with OpenXR MNDs xdev spaces.

Mr-Zero88: Done.

Mr-Zero88:
> Looks like the tests need to be updated?

Ah, I missed the fakexr entry while adding platform_info. The tests itself should be fine.

Issue #365, opened 2026-05-31
[Bug]: Fallout 4 VR default controls for Oculus Touch have incorrect/unmapped bindings for buttons and sticks


SeongGino:
### Operating System

Arch Linux

### OpenXR Runtime

WiVRn

### Describe the issue.

On a Quest 2 setup (may also apply to Rift/S, Quest 1/3(S)) with Oculus Touch controllers, FO4VR is *supposed* to have the following controls, as indicated by the ingame menu->Help section:

<img width="1116" height="1256" alt="Image" src="https://github.com/user-attachments/assets/0f0bee95-dc19-4736-9ac5-c177401c6929" />
<img width="1116" height="1256" alt="Image" src="https://github.com/user-attachments/assets/3415d642-cd5d-447e-acef-e0c9b761522d" />
<img width="1116" height="1256" alt="Image" src="https://github.com/user-attachments/assets/d318125c-e217-4de4-bb8d-7034fd197fd8" />

Out of these, the Left/Right hand bottom face buttons (A/X) and stick clicks don't seem to be mapped, so the player cannot use interact menus, can't access the favorites wheel. can't scroll back/forwards through main sections in the PipBoy menu (depending on handedness), and can't sprint; only the triggers, grip buttons, basic stick directions/flicks, and the top face buttons (Y/B) work.

###### I'm not sure why there's an errant mention of what looks like a Vive Wand touchpad for sheathing, when the rest is correctly describing Oculus Touch...

I want to say this might also in part apply to Skyrim VR as it would be using the same base engine/share a porting team, but I don't currently own it to test.

As this game(/pair) uses a VR port of Bethesda's Creation Engine instead of the typical Unity or Unreal, there is no such `StreamingAssets/SteamVR` bindings folder, nor does it really use modern Steam Input bindings, so it's hard-coded to specific controls based on controller.

### Steps to reproduce

1. Start up WiVRn with an Oculus Quest
2. Start Fallout 4 VR (preferably using Proton Experimental)
3. Load a save past the introduction sequence (at least after reawakening in the Vault, preferably outside)
4. Try to access the favorites wheel, sprint/use teleport movement, interact with/loot objects/containers, or access the PipBoy menu categories as the game describes via tutorials or the in-game Help.

### Please confirm you have reproduced this on the latest commit.

- [x] I have reproduced this on the latest commit

BiggestGoos:
[...]

PR #430, opened 2026-09-19
input: report HMD identity from the controller interaction profile

mkopec:
Heads up: This is vibe coded using Claude Opus 5 and my rust is quite terrible, so it's hard for me to attest to the quality of the code. That said, the logic seems sound to me. If LLM contributions aren't desired, feel free to close this immediately (I couldn't find precedent either way).

This is an attempt to fix and #365 and #394. per the suggestion this comment: https://github.com/Supreeeme/xrizer/pull/394#issuecomment-5344061917

Games like Fallout 4 VR read the HMD's TrackingSystemName_String once, right after init, two changes are needed to support this:

1. Give the HMD an identity. Values are taken from Valve's Steam Link driver config.
2. Know the profile in time. Spin up a throwaway session with a dummy action set bound to the grip pose of every supported profile, cache the bound profiles and tear the session down again. The HMD identity is taken from that cache.

Tested, Fallout 4 VR now uses the correct controller bindings with a Quest 3 via WiVRn. Other controllers or headsets weren't tested. All tests pass.


CrashTD:
Tested on Skyrim VR with a big modlist (Mad God Overhaul, ~1800 mods), Quest 3 over WiVRn, CachyOS, proton-cachyos-native.

Without the PR: only trigger and grip work, in menus and in game – sticks and face buttons are dead. Vanilla Skyrim VR on the same runtime is fine, so it's something in the modlist reading the HMD identity, not the bare engine. Everything also works with OpenComposite instead of xrizer.

With this PR (built on 0989a7f, reports v0.5-25-g3bf37d7) it works – with one catch: the controls stay dead until I press trigger or grip once. After that first input, sticks and buttons work for the rest of the session. That lines up with the fallback: until an interaction profile exists, the properties still return `<unknown>`, and whatever reads them in the modlist seems to read early and only once.

